Issue #40, opened 2022-09-05
Fix test for pool announcement


afritzler:
**Describe the bug**
Currently we are creating `VolumePools` manually in the volume pool reconciler. This should be removed and instead we should test that the `birthCry` successfully created a `VolumePool` as configured via the `VolumeReconciler` fields. All subsequent tests should be based on this pool.

PR #46, opened 2022-09-06
Reconcile only `VolumePool` created by `VolumePoolReconciler`

### LINKED ISSUE #42: Only reconcile `VolumePools` which have been created by the `VolumePool` reconciler
# Summary

Currently our `VolumePoolReconciler` reconciles every `VolumePool` resource in the cluster. We should this restrict only to the ones we have created in the startup process of the manager. That way we can prevent unwanted behaviour/side-effects when users create a custom `VolumePool` in the cluster.

In future we need to think about what it means to manage multiple `VolumePools` and how this should be supported by the `cephlet`.

lukasfrank:
# Proposed Changes

- Check if `VolumePool` was created by `VolumePoolReconciler` by comparing pool name
- Adapted tests

Fixes 
#42 
#40

### DIFF
--- a/controllers/volumepool_controller.go
+++ b/controllers/volumepool_controller.go
@@ -81,6 +81,12 @@ func (r *VolumePoolReconciler) reconcileExists(ctx context.Context, log logr.Log
 	if !pool.DeletionTimestamp.IsZero() {
 		return r.delete(ctx, log, pool)
 	}
+
+	if pool.Name != r.VolumePoolName {
+		log.V(1).Info("Skipped pool, since it was not created by controller", "pool", client.ObjectKeyFromObject(pool))
+		return ctrl.Result{}, nil
+	}
+
 	return r.reconcile(ctx, log, pool)
 }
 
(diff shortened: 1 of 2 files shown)


controllers/volumepool_controller.go
```diff
@@ -81,6 +81,12 @@ func (r *VolumePoolReconciler) reconcileExists(ctx context.Context, log logr.Log
 	if !pool.DeletionTimestamp.IsZero() {
 		return r.delete(ctx, log, pool)
 	}
+
+	if pool.Name != r.VolumePoolName {
+		log.V(1).Info("Skipped pool, since it was not created by controller", "pool", client.ObjectKeyFromObject(pool))
```
afritzler, line 86:
```suggestion
		log.V(1).Info("Skipping VolumePool, since it is not owned by us")
```

afritzler: No need to print the pool as it is automatically done by the `VolumePoolReconciler` logger. (the main `For` type is always printed in the structured log)

afritzler (changes requested):

afritzler (approved):

PR #50, opened 2022-09-07
Stream image from registry

### LINKED ISSUE #45: Stream data to block device instead of copying the artefacts
# Summary

Currently we are copying the OCI image artefacts into the populator pod before writing them to the block device. This might cause a problem if the OCI image is to large and might cause to `Pod` to run out of disk space.

## Proposed solution

Use `io.Reader`/`io.Writer` to stream the content from the remote location directly to the block device.

lukasfrank:
# Proposed Changes

- Stream image directly to block device
- Remove not needed code

Fixes 
#45


cmd/populator/populator.go
```diff
@@ -34,60 +34,75 @@ const (
...
-func populate(log logr.Logger, storePath string, ref string, devicePath string) {
+func populate(log logr.Logger, ref string, devicePath string) error {
 	ctx := ctrl.SetupSignalHandler()
 	log.Info("Starting image population")
 
 	reg, err := remote.DockerRegistry(nil)
 	if err != nil {
 		log.Error(err, "Failed to initialize registry")
-		os.Exit(1)
+		return err
```
afritzler, line 60: Please return an `fmt.Errorf` here.

cmd/populator/populator.go
```diff
@@ -34,60 +34,75 @@ const (
...
-	flag.StringVar(&storePath, "storePath", "/tmp", "Location of the local image store")
 	opts := zap.Options{
 		Development: true,
 	}
 	opts.BindFlags(goflag.CommandLine)
 	flag.Parse()
 
-	populate(zap.New(zap.UseFlagOptions(&opts)), storePath, image, devicePath)
+	if err := populate(zap.New(zap.UseFlagOptions(&opts)), image, devicePath); err != nil {
+		os.Exit(1)
```
afritzler, line 49: And here print the error via `log.Error`

afritzler (changes requested):

cmd/populator/populator.go
```diff
@@ -45,31 +46,30 @@ func main() {
...
-func populate(log logr.Logger, ref string, devicePath string) error {
+func populate(log logr.Logger, ref string) error {
 	ctx := ctrl.SetupSignalHandler()
 	log.Info("Starting image population")
 
 	reg, err := remote.DockerRegistry(nil)
 	if err != nil {
-		log.Error(err, "Failed to initialize registry")
-		return err
+		return fmt.Errorf("failed to initialize registry")
```
afritzler, line 62: You want to wrap the error here via `fmt.Errorf("....: %w", err)`

afritzler: Same applies for the other error returns

afritzler (changes requested):

cmd/populator/populator.go
lukasfrank: copy&paste 🤦‍♂️

afritzler (approved):

PR #72, opened 2022-10-05
Set `VolumeLimits` using the `Ceph API` 

lukasfrank:
# Proposed Changes

- Attach the `Limit Information` to the ceph volume by using the ceph dashbard api
- Limits will be fetched through the meta data channel by the clients


pkg/ceph/client.go
```diff
@@ -0,0 +1,202 @@
...
+	http       *http.Client
+	rookConfig *rook.Config
+}
+
+func (t tokenSource) Token() (*oauth2.Token, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+
+	secret := &corev1.Secret{}
+	if err := t.Get(context.TODO(), types.NamespacedName{Namespace: t.rookConfig.Namespace, Name: t.rookConfig.DashboardSecretName}, secret); err != nil {
```
afritzler, line 76: Please use the context defined above.

afritzler (changes requested):

afritzler (approved):

PR #94, opened 2022-10-25
Watch `VolumePool`

lukasfrank:
# Proposed Changes

- Reconcile all `Volumes` with a reference to a `VolumePool` once it gets available

### DIFF
--- a/controllers/volume_controller.go
+++ b/controllers/volume_controller.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


controllers/volume_controller.go
```diff
@@ -483,6 +500,27 @@ func (r *VolumeReconciler) getMonitorList(ctx context.Context) ([]string, error)
 	return monitors, nil
 }
 
+func (r *VolumeReconciler) findObjectsForVolumePool(pool client.Object) []reconcile.Request {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
afritzler, line 504: Is there a reason why you used a time out context here? Shouldn't 
```	
ctx, cancel := context.WithCancel(context.Background())
```

be enough here?

afritzler (changes requested):

controllers/volume_controller.go
lukasfrank: The timeout is used to restrict the `if err := r.List(ctx, volumes, &client.ListOptions{` statement.
With a `context.WithCancel` context the request can dangle around forever and the `cancel` callback will not be called.

afritzler: What if the amount of `Volume`s is large and it takes longer than 1s to `List` them?

lukasfrank: If the context closes or an other error is thrown by `r.List` an empty list is returned and no `Volume` is being queued. But yes: Maybe the timeout is too small. Do you have there a opinion what value to use?

PR #143, opened 2022-12-12
Use same size for Snapshot as rootFS image layer

### LINKED ISSUE #126: snapshot size should equals to image size
currently the size of the snapshot is the size of the first image requested. if any following image using this snapshot is smaller we have various error -> not supported by ceph but collapses in itself 

make sure if we have a disk that depends on a snapshot (with the image capability) it must have minimum the size of the image (just to avoid any further bugs like this)

lukasfrank:
# Proposed Changes

Fixes #126


controllers/volume_controller.go
```diff
@@ -350,6 +354,37 @@ func (r *VolumeReconciler) handleImagePopulation(ctx context.Context, log logr.L
 }
 
 func (r *VolumeReconciler) createSnapshot(ctx context.Context, log logr.Logger, volume *storagev1alpha1.Volume) error {
+	reg, err := remote.DockerRegistry(nil)
```
afritzler, line 357: You might want to put the size calculation into a helper function.

afritzler (changes requested):

afritzler (approved):

PR #196, opened 2023-02-28
Remove `ceph-csi`

### LINKED ISSUE #187: Remove `ceph-csi` dependency
# Summary

Currently, the `cephlet` uses PVCs provisioned by the `ceph-csi`  in order to back the onmetal `Volume`. This dependency should be removed since it simplifies the further development. 

## Basic example

Current flow
1. Volume
2. `cephlet` creats PVC
3. PVC gets provisioned by csi 

should be simplified by
1. Volume
2. `cephlet` talks directly to `ceph` 

## Motivation

Benefits: 
- remove not needed components / abstractions 
- simplify image population 
- simplify storage encryption

lukasfrank:
# Proposed Changes

-
-
-

Fixes #187


Dockerfile
```diff
@@ -1,8 +1,143 @@
 # Build the manager binary
-FROM --platform=$BUILDPLATFORM golang:1.20.1 as builder
+FROM --platform=$BUILDPLATFORM debian:bookworm-slim as builder
```
adracus, line 2: Why do we setup golang entirely on our own here? The default `golang:1.20.1` points to a version of Debian bullseye, which is the latest version of debian. bookworm is still in testing phase, isn't it?

Dockerfile
```diff
@@ -52,16 +186,62 @@ USER 65532:65532
...
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libz.so.1 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libbz2.so.1.0 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/liblzma.so.5 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libzstd.so.1 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libnl-route-3.so.200 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libnl-3.so.200 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libselinux.so.1 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libpthread.so.0 \
+/lib/${LIB_DIR_PREFIX}-linux-gnu/libpcre2-8.so.0 /lib/${LIB_DIR_PREFIX}-linux-gnu
+COPY --from=ori-volume-builder /lib64/ld-linux-${LIB_DIR_PREFIX_MINUS}.so.2 /lib64/
```
adracus, line 240: I think it might make sense to have a 'validation' image in between validating that all shared libraries are present.
For reference, we do the same in `onmetal-csi-driver`:
https://github.com/onmetal/onmetal-csi-driver/blob/fddb2662974d1db48f8efdb3ff878c155431060a/Dockerfile#L109

controllers/bucketpool_controller.go
```diff
@@ -42,8 +42,7 @@ import (
 )
 
 const (
-	bucketPoolFinalizer        = "cephlet.onmetal.de/bucketpool"
-	bucketPoolSecretAnnotation = "ceph-client-secret-name"
+	bucketPoolFinalizer = "cephlet.onmetal.de/bucketpool"
```
adracus, line 45: Shouldn't the group be `cephlet.api.onmetal.de` since `onmetal.de` is not under the `api` group?

ori/volume/server/server.go
```diff
@@ -17,66 +17,77 @@ package server
...
+	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
-var scheme = runtime.NewScheme()
+const (
+	OsImage  = true
+	RbdImage = false
```
adracus, line 34: Why do we have these constants?

ori/volume/server/server.go
```diff
@@ -17,66 +17,77 @@ package server
...
-	volumeClassSelector client.MatchingLabels
+	return nil
+}
+
+func (s *Server) Release(volumeName string) {
+	s.inProgressLock.Lock()
+	defer s.inProgressLock.Unlock()
+
+	s.inProgress.Delete(volumeName)
 }
```
adracus, line 84: The server should not be the responsible entity for locking.

ori/volume/server/server.go
```diff
@@ -126,52 +123,39 @@ var _ ori.VolumeRuntimeServer = (*Server)(nil)
...
+		inProgress:             map[string]sets.Empty{},
+		inProgressLock:         sync.Mutex{},
 	}, nil
 }
 
-func (s *Server) getManagedAndCreated(ctx context.Context, name string, obj client.Object) error {
-	key := client.ObjectKey{Namespace: s.namespace, Name: name}
-	if err := s.client.Get(ctx, key, obj); err != nil {
-		return err
+func ReadVolumeClasses(path string) (map[string]ori.VolumeClass, error) {
```
adracus, line 144: 1. The volume classes read here are actually part of the configuration of the `cephlet`. I'd recommend extending this already to read a config file where one value (maybe currently the only one) is the list of volume classes.
2. The returned value should be the structure read from the file. Transformations like the one to get the classes by name should be done by the caller afterwards. E.g. right now, your function could hide the truth that there are volume classes with duplicate names.

ori/volume/server/server.go
```diff
@@ -126,52 +123,39 @@ var _ ori.VolumeRuntimeServer = (*Server)(nil)
...
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read volume class json file (%s): %w", path, err)
```
adracus, line 147:
```suggestion
		return nil, fmt.Errorf("error reading config file at %q: %w", path, err)
```
Using `%q` quotes the string. We're also not sure whether it's an actual JSON file at this point.

ori/volume/server/server.go
```diff
@@ -126,52 +123,39 @@ var _ ori.VolumeRuntimeServer = (*Server)(nil)
...
-		if err != nil {
-			return err
-		}
 
-		return apierrors.NewNotFound(schema.GroupResource{
-			Group:    gvk.Group,
-			Resource: gvk.Kind, // Yes, kind is good enough here
-		}, key.Name)
+	var classList []ori.VolumeClass
+	if err := json.Unmarshal(content, &classList); err != nil {
```
adracus, line 151: Here we could use the `yaml.NewYAMLOrJSONDecoder` as in k8s, allowing the config to be either json or yaml.

ori/volume/server/volume.go
```diff
@@ -16,235 +16,143 @@ package server
...
-	userIDKey  = "userID"
-	userKeyKey = "userKey"
-
-	pvPoolKey               = "pool"
-	monitorConfigMapDataKey = "csi-cluster-config-json"
-	pvImageNameKey          = "imageName"
-	pvRadosNamespaceKey     = "radosNamespace"
+	// worldwide number key
+	// to use WWN Company Identifiers, set wwnPrefix to Private "1100AA"
+	wwnPrefix string = ""
```
adracus, line 32: Why this constant if it's `""` by default? If it should be overridable, it should be a `var` / a config field of an entity that uses it.

ori/volume/server/volume.go
```diff
@@ -16,235 +16,143 @@ package server
...
-	wwn := wwnPrefix
+type Volume struct {
+	Name string
 
-	// use UUIDv4, because this will generate good random string
-	wwnUUID, err := uuid.NewRandom()
-	if err != nil {
-		return "", fmt.Errorf("failed to generate UUIDv4 for WWN: %w", err)
-	}
+	Bytes uint64
```
adracus, line 48: What is meant with `Bytes` here? Is it actually the size in bytes? If so, I'd recommend naming it `Size` (e.g. also the same name used in OCI descriptors).

ori/volume/server/volume.go
```diff
@@ -16,235 +16,143 @@ package server
...
-	return wwn[:16], nil
+	Image *PopulationImage
 }
 
-func (s *Server) convertOnmetalVolumeState(state corev1.PersistentVolumeClaimPhase) (ori.VolumeState, error) {
-	if state, ok := onmetalVolumeStateToORIState[state]; ok {
-		return state, nil
-	}
-	return 0, fmt.Errorf("unknown onmetal volume state %q", state)
+type PopulationImage struct {
```
adracus, line 57: Something about the naming here `PopulationImage` / `Image` seems a bit off - let's discuss about this.

ori/volume/server/volume.go
```diff
@@ -16,235 +16,143 @@ package server
...
+	Name  string
+	Bytes uint64
```
adracus, line 59: Same as above comments about naming.

ori/volume/server/volume.go
```diff
@@ -16,235 +16,143 @@ package server
...
-	credentials, ok := cephClientSecret.Data[s.rookClientName]
-	if !ok {
-		return nil, fmt.Errorf("secret %s does not contain data key %s", client.ObjectKeyFromObject(cephClientSecret), s.rookPoolName)
-	}
-
-	return credentials, nil
+func (s *Server) createOriObjectMetadata(image *Image) (*orimeta.ObjectMetadata, error) {
+	return &orimeta.ObjectMetadata{
+		Id:         image.Name,
+		Generation: 0,
```
adracus, line 136: How is the `Generation` / `DeletedAt` being handled?

ori/volume/server/volume_list.go
```diff
@@ -18,64 +18,39 @@ import (
...
-	onmetalVolume, err := s.getAggregateCephVolume(ctx, id)
+	found, err := s.provisioner.GetCephImage(ctx, id, cephImage)
 	if err != nil {
 		return nil, err
 	}
 
-	return s.convertAggregateCephVolume(ctx, onmetalVolume)
+	if !found {
+		log.V(1).Info("Inconsistent state: Failed to find image", "id", id)
+		return nil, fmt.Errorf("failed to find image")
```
adracus, line 39: Why is this 'inconsistent'? What if an id actually does not exist? Shouldn't we report this back?

pkg/populate/copy.go
```diff
@@ -0,0 +1,60 @@
...
+	n, err = r.r.Read(b)
+	r.count += int64(n)
+	if err == io.EOF {
+		r.end = time.Now()
+	}
+
+	return
+}
+
+func (r *rate) Rate() (n int64, d time.Duration) {
```
adracus, line 49: What if `Rate` gets called before `r.start` is set?

pkg/round/size.go
```diff
@@ -0,0 +1,51 @@
...
+	// KB - KiloByte size
+	KB = 1000
+	// KiB - KibiByte size
+	KiB = 1024
+)
+
+// OffBytes converts roundoff the size
+// 1.1Mib will be round off to 2Mib same for GiB
+// size less than 1MiB will be round off to 1MiB.
+func OffBytes(bytes uint64) uint64 {
```
adracus, line 39: Can't we use Kubernetes' `resource` package (e.g. `resource.NewQuantity`)?

ori/volume/server/volume_delete.go
```diff
@@ -18,27 +18,58 @@ import (
...
+
+func (s *Server) deleteCephImage(ctx context.Context, log logr.Logger, volumeId string) (retErr error) {
+
+	volumeName, found, err := s.findNameFromId(ctx, volumeId)
+	if err != nil {
+		return fmt.Errorf("unable to find volumeName from id: %w", err)
+	}
+	if !found {
+		log.V(1).Info("No mapping found: Already gone.", "imageId", volumeId)
+		return nil
```
adracus, line 48: We should return a proper error if we could not find an image.

ori/volume/server/volume_delete.go
```diff
@@ -18,27 +18,58 @@ import (
...
+	}
+
+	log.V(2).Info("Try to acquire lock for volume", "volumeName", volumeName)
+	if err := s.Lock(volumeName); err != nil {
+		return fmt.Errorf("unable to acquire lock: %w", err)
+	}
+	defer s.Release(volumeName)
```
adracus, line 55: `Lock`ing / `Release`ing should be done at the server level.

lukasfrank: `s.Lock`/`s.Release` operates on the server

Dockerfile
lukasfrank: `libcephfs-dev`  `librbd-dev` `librados-dev` is not present in `bullseye` therefore the other option would be to build this three libs from scratch. Once the new `go image` uses `bookworm` the manual go building stuff will be removed and the rest stays the same.

ori/volume/server/server.go
lukasfrank: I do assume, that a `ori.VolumeClass` List is being passed, it only has `json` tags

pkg/populate/copy.go
lukasfrank: Rate will calculate diff between `time.Now()` and a uninitialized  `time.Time` (without a panic)

pkg/populate/populate.go
```diff
@@ -0,0 +1,78 @@
...
+			case <-ticker.C:
+				log.Info("Populating", "rate", rater.String())
+			case <-done:
+				return
+			}
+		}
+	}()
+	defer func() { done <- true }()
+
+	buffer := make([]byte, bufferSize)
```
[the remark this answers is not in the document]
lukasfrank: I'm not sure if it's worth it. This code only get's executed only once in a while (if a new image+tag) gets used. I think it will not produce a lot memory pressure. This simpler solution will have most of the time a smaller memory footprint since we hand back the buffer.

pkg/round/size.go
lukasfrank: We can but since we are dealing all the time with "plain" byte count's it's more handy to not convert to a `ressource`. Furthermore you need first to check for the `Scale` in order to pass it to the `RoundUp` function e.g. `RoundUp(resource.Giga)`

adracus: Better question: Where is this used / required? Maybe the caller could make better use of the `resource` package.

lukasfrank: This is used to provision rbd images without fragmentation.

.github/workflows/publish-docker.yml
```diff
@@ -73,7 +69,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           context: .
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64 #,linux/arm64
```
adracus, line 72: Is it not possible anymore to build `linux/arm64` after this change?

ori/volume/server/volume.go
```diff
@@ -16,235 +16,107 @@ package server
...
 )
 
-type AggregateCephVolume struct {
-	Pvc          *corev1.PersistentVolumeClaim
-	ImagePvc     *corev1.PersistentVolumeClaim
-	VolumeClass  *storagev1alpha1.VolumeClass
-	AccessSecret *corev1.Secret
+type PopulationImage struct {
+	Name  string
+	Bytes uint64
```
adracus, line 36: This should be called `Size`.

ori/volume/server/server.go
```diff
@@ -17,66 +17,72 @@ package server
...
+	inProgressLock sync.Mutex
+
+	VolumeNameLabelName    string
+	AvailableVolumeClasses map[string]ori.VolumeClass
 
-	namespace string
+	syncPopulation singleflight.Group
+}
+
+func (s *Server) lock(volumeName string) error {
```
adracus, line 61: Why does this return an `error`? It has to prevent multiple operations from operating on the same volume, yes, but it shouldn't error and rather let the subsequent callers wait until the previous lock has been released. You could (probably, don't pin me on that) achieve that using `sync.Cond` / to a certain degree even with the `singleflight.Group` (depending on impementation ofc).

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,167 @@ package server
...
-		ImagePvc:    imagePvc,
-		VolumeClass: volumeClass,
-	}, nil
 }
 
-func (s *Server) handleImagePopulation(ctx context.Context, log logr.Logger, volume *AggregateCephVolume) error {
-	if volume.ImagePvc == nil {
-		return nil
-	}
+func (s *Server) prepareOSImage(ctx context.Context, log logr.Logger, osImageName string) (osImageId string, retErr error) {
```
adracus, line 92: I'd recommend moving the functionality of downloading / preparing base images to another component. That component then can, given an image id, do the singleflight downloading.

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,167 @@ package server
...
+	if found {
+		return osImageId, nil
 	}
 
-	volumeSizeBytes := volume.Pvc.Spec.Resources.Requests.Storage().Value()
-	if volumeSizeBytes < snapshot.Status.RestoreSize.Value() {
-		log.Info(fmt.Sprintf("Requested volume size %d is less than the size %d for the source snapshot", volumeSizeBytes, snapshot.Status.RestoreSize.Value()), "snapshotName", snapshot.Name)
-		//r.Eventf(volume, corev1.EventTypeWarning, ReasonVolumeSizeToSmall, "Requested volume size %d is less than the size %d for the source snapshot %s", volumeSizeBytes, snapshot.Status.RestoreSize.Value(), snapshot.Name)
+	osImageId = s.idGen.Generate()
+	if err := s.provisioner.CreateOsImage(ctx, osImageName, osImageId); err != nil {
```
adracus, line 106: Why do we need to generate an id for an os image? The unique identifier of an image should be its digest (coming from `Image.Digest()`). The downloading entity should be able to pull an image by the image name and return the digest, additional mappings limit re-use / make the whole flow more complex.

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,167 @@ package server
...
-		return fmt.Errorf("error populate image: %w", err)
+		populationImageId = osImageId.(string)
 	}
 
-	c, cleanup := s.setupCleaner(ctx, log, &retErr)
-	defer cleanup()
+	class, found := s.AvailableVolumeClasses[volume.Spec.Class]
+	if !found {
+		return nil, fmt.Errorf("not supported volume class: %s", volume.Spec.Class)
+	}
```
adracus, line 147: Shouldn't this check actually happen before provisioning?

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,167 @@ package server
...
-	if err := s.handleImagePopulation(ctx, log, volume); err != nil {
-		return fmt.Errorf("error populate image: %w", err)
+		populationImageId = osImageId.(string)
 	}
 
-	c, cleanup := s.setupCleaner(ctx, log, &retErr)
-	defer cleanup()
+	class, found := s.AvailableVolumeClasses[volume.Spec.Class]
+	if !found {
+		return nil, fmt.Errorf("not supported volume class: %s", volume.Spec.Class)
```
adracus, line 146:
```suggestion
		return nil, fmt.Errorf("unsupported volume class: %s", volume.Spec.Class)
```

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,167 @@ package server
...
-		//r.Eventf(volume, corev1.EventTypeWarning, ReasonVolumeSizeToSmall, "Requested volume size %d is less than the size %d for the source snapshot %s", volumeSizeBytes, snapshot.Status.RestoreSize.Value(), snapshot.Name)
+	osImageId = s.idGen.Generate()
+	if err := s.provisioner.CreateOsImage(ctx, osImageName, osImageId); err != nil {
+		return "", fmt.Errorf("unable to create os image: %w", err)
 	}
+	c.Add(func(ctx context.Context) error {
+		log.V(2).Info("Delete os image")
+		return s.provisioner.DeleteOsImage(ctx, osImageName, osImageId)
+	})
+	c.Reset()
```
adracus, line 113: Why is this call here?

pkg/populate/populate.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/go-logr/logr"
+	onmetalimage "github.com/onmetal/onmetal-image"
+	"github.com/onmetal/onmetal-image/oci/remote"
+)
+
+func ResolveImage(ctx context.Context, log logr.Logger, imageName string) (*onmetalimage.Image, error) {
+	log.V(1).Info("Resolving image", "name", imageName)
+	reg, err := remote.DockerRegistry(nil)
```
adracus, line 30: We shouldn't instantiate the `remote.DockerRegistry` for each image resolution call - some entity should store the result of this. Maybe another thing to move to a potential image fetching entity?

adracus (changes requested):

ori/volume/server/volume.go
```diff
@@ -16,235 +16,107 @@ package server
...
 		},
 	}, nil
 }
-
-func (s *Server) convertVolumeResources(resources corev1.ResourceList) (*ori.VolumeResources, error) {
-	storage := resources.Storage()
-	if storage.IsZero() {
-		return nil, fmt.Errorf("volume does not specify storage resource")
+func (s *Server) getOriVolumeAccess(ctx context.Context, log logr.Logger, image *CephImage) (*ori.VolumeAccess, error) {
+	user, key, err := s.provisioner.FetchAuth(ctx)
```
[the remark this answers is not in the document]
lukasfrank: `s.provisioner.FetchAuth(ctx)` selects a `ceph client` created by `rook` and returns the credentials. This is a pre-condition to be created in advance.

ori/volume/server/server.go
lukasfrank: I think it's fine to return an error because of the following reasons: 
- we are locking based on a ceph image id. In case of volume creation it will only return an error if we have selected a duplicate  UUID
- in case of operating on the same image at the same time (like deleting a `Volume` which is not fully provisioned): that's not allowed and we can do an early return. The caller should do a retry

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,167 @@ package server
...
-		log.Info(fmt.Sprintf("Requested volume size %d is less than the size %d for the source snapshot", volumeSizeBytes, snapshot.Status.RestoreSize.Value()), "snapshotName", snapshot.Name)
-		//r.Eventf(volume, corev1.EventTypeWarning, ReasonVolumeSizeToSmall, "Requested volume size %d is less than the size %d for the source snapshot %s", volumeSizeBytes, snapshot.Status.RestoreSize.Value(), snapshot.Name)
+	osImageId = s.idGen.Generate()
+	if err := s.provisioner.CreateOsImage(ctx, osImageName, osImageId); err != nil {
+		return "", fmt.Errorf("unable to create os image: %w", err)
 	}
+	c.Add(func(ctx context.Context) error {
+		log.V(2).Info("Delete os image")
+		return s.provisioner.DeleteOsImage(ctx, osImageName, osImageId)
+	})
```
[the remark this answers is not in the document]
lukasfrank: The cleanup is not about deleting `unused` os images but rather about deleting a "corrupted/not fully provisioned" image.

.github/workflows/publish-docker.yml
lukasfrank: Correct

ori/volume/server/volume.go
```diff
@@ -16,235 +16,107 @@ package server
...
+	Annotations map[string]string
+	Labels      map[string]string
+	Generation  int64
+	CreatedAt   time.Time
+	DeletedAt   time.Time
+
+	Wwn                string
+	Pool               string
+	Size               uint64
+	PopulatedImageName string
```
adracus, line 50: This should be called `Image`.

ori/volume/server/volume.go
```diff
@@ -16,235 +16,107 @@ package server
...
+	Class              string
+	State              ori.VolumeState
```
adracus, line 52: `State` should be specific to your implementation and not use `ori` types. Your internal types have to function without `ori`.

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,132 @@ package server
...
-	}
+	return nil
+}
 
-	size := resource.NewQuantity(rootFSLayer.Descriptor().Size, resource.BinarySI)
-	if size == nil {
-		return nil, fmt.Errorf("failed to get size of rootFS layer")
+func (s *Server) validateSize(volume *ori.Volume, image *PopulationImage) error {
+	if image != nil && volume.Spec.Resources.StorageBytes <= image.Size {
+		return fmt.Errorf("volume size smaller than image")
```
adracus, line 36: Please include the expected and actual size in the error message.

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,132 @@ package server
...
-					APIGroup: pointer.String(storagev1alpha1.SchemeGroupVersion.String()),
-					Kind:     "Volume",
-					Name:     pvc.Name,
-				},
-			},
-		}
-		metautils.SetLabel(imagePvc, volumev1alpha1.ManagerLabel, volumev1alpha1.VolumeCephletManager)
-
+	log.V(2).Info("Resolving os image")
+	img, err := s.populator.ResolveImage(ctx, imageName)
```
adracus, line 49: Validating should not include resolving the image. A create with an image that is not yet available should succeed since e.g. the push to the registry might happen at a later point in time. For comparison, see e.g. image pulling in pods in k8s. A pod's image might not be there upon creation of the `corev1.Pod`, but the container runtime retries indefinitely (unless the pod is deleted in the meantime).

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,132 @@ package server
...
+	log.V(2).Info("Validated image")
 
-	volumeSizeBytes := volume.Pvc.Spec.Resources.Requests.Storage().Value()
-	if volumeSizeBytes < snapshot.Status.RestoreSize.Value() {
-		log.Info(fmt.Sprintf("Requested volume size %d is less than the size %d for the source snapshot", volumeSizeBytes, snapshot.Status.RestoreSize.Value()), "snapshotName", snapshot.Name)
-		//r.Eventf(volume, corev1.EventTypeWarning, ReasonVolumeSizeToSmall, "Requested volume size %d is less than the size %d for the source snapshot %s", volumeSizeBytes, snapshot.Status.RestoreSize.Value(), snapshot.Name)
+	if err := s.validateSize(oriVolume, validatedImage); err != nil {
+		return fmt.Errorf("invalid volume size: %w", err)
 	}
+	log.V(2).Info("Validated size")
```
adracus, line 86: Now that I see that the size depends on the image, you cannot do this validation at creation time. You'll have to add this as a runtime error / mark the volume as state == Errored or sth. like that when actually trying to pull.

pkg/populator/populator.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func New(log logr.Logger, provisioner Provisioner) (*populator, error) {
+	var err error
+	p := &populator{
+		log:         log,
+		provisioner: provisioner,
+		populated:   map[string]string{},
+	}
+	p.registry, err = remote.DockerRegistry(nil)
```
adracus, line 43: This should be passed in (e.g. via an `Options` struct) and not statically assigned.

pkg/vcr/registry.go
```diff
@@ -0,0 +1,75 @@
...
+	"io"
+	"os"
+
+	ori "github.com/onmetal/onmetal-api/ori/apis/volume/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func LoadVolumeClasses(reader io.Reader) ([]ori.VolumeClass, error) {
+	classList := []ori.VolumeClass{}
+	if err := yaml.NewYAMLOrJSONDecoder(reader, 100).Decode(&classList); err != nil {
```
adracus, line 28: We should have a more sensible default (`4096`) here - see e.g. `controller-utils/unstructuredutils`

pkg/vcr/registry.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	ori "github.com/onmetal/onmetal-api/ori/apis/volume/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func LoadVolumeClasses(reader io.Reader) ([]ori.VolumeClass, error) {
+	classList := []ori.VolumeClass{}
```
adracus, line 27: Empty slices should be declared using the zero value, i.e.

```go
var classList []ori.VolumeClass
```

pkg/vcr/registry.go
```diff
@@ -0,0 +1,75 @@
...
+func LoadVolumeClassesFile(filename string) ([]ori.VolumeClass, error) {
+	file, err := os.Open(filename)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open volume class file (%s): %w", filename, err)
+	}
+
+	return LoadVolumeClasses(file)
+}
+
+func NewVolumeClassRegistry(classes []ori.VolumeClass) (*vcr, error) {
```
adracus, line 44: The return type should either be an interface or be exported.

pkg/populator/populator.go
```diff
@@ -0,0 +1,160 @@
...
+	registry    *remote.Registry
+	provisioner Provisioner
+
+	populated map[string]string
+	lock      sync.RWMutex
+
+	populating singleflight.Group
+}
+
+func (p *populator) isIdCached(imageName string) (string, bool) {
```
adracus, line 63: This caching should be removed. There should be an efficient lookup in the existing images.

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
-	img, err := reg.Resolve(ctx, imageName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve image ref in registry: %w", err)
-	}
-
-	layers, err := img.Layers(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get layers for image: %w", err)
-	}
+func (s *Server) getAggregatedImage(ctx context.Context, log logr.Logger, volume *ori.Volume) (*aggregatedImage, error) {
```
adracus, line 37: If this is used to compute the configuration (i.e. required values for the internal API), this should be called `getCephVolumeConfig`. The return type should then be called `CephVolumeConfig`.

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
+		snapshotRef = pointer.String(img.Descriptor().Digest.String())
+		snapshot = &api.Snapshot{
+			Metadata: api.Metadata{
+				ID: img.Descriptor().Digest.String(),
+			},
+			Source: api.SnapshotSource{
+				OnmetalImage: image,
+			},
+		}
+		labels[SnapshotNameLabel] = image
```
adracus, line 91: Why are you writing into user-defined labels?

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
+	labels := volume.Metadata.Labels
+	if labels == nil {
+		labels = map[string]string{}
 	}
 
-	var imagePvc *corev1.PersistentVolumeClaim
-	if volume.Spec.Image != "" {
-		imageName := GetSanitizedImageName(volume.Spec.Image)
-		apiutils.SetImageAnnotation(pvc, volume.Spec.Image)
+	labels[VolumeClassLabel] = volume.Spec.Class
```
adracus, line 58: Why are you writing into user-defined labels?

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
-		},
-		Spec: snapshotv1.VolumeSnapshotSpec{
-			Source: snapshotv1.VolumeSnapshotSource{
-				PersistentVolumeClaimName: &volume.ImagePvc.Name,
+	return &aggregatedImage{
+		image: &api.Image{
+			Metadata: api.Metadata{
+				ID:          s.idGen.Generate(),
+				Annotations: volume.Metadata.Annotations,
+				Labels:      labels,
```
adracus, line 99: There has to be a separation between labels & annotations required by the `ori` server implementation and labels and annotations specified by the `CreateVolume` call.

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
+			},
+			Spec: api.ImageSpec{
```
adracus, line 101: Why is this rounded here? Shouldn't it stick to the users specification? Also, if it's rounded here and you return it later as the rounded size, the calling entity will have problems since the specified size differs from the actual size.

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
-func (s *Server) handleImagePopulation(ctx context.Context, log logr.Logger, volume *AggregateCephVolume) error {
-	if volume.ImagePvc == nil {
-		return nil
-	}
-
-	snapshot := &snapshotv1.VolumeSnapshot{}
-	if err := s.client.Get(ctx, client.ObjectKeyFromObject(volume.ImagePvc), snapshot); err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("unable to get snapshot: %w", err)
+		imageSize := uint64(onmetalImage.RootFS.Descriptor().Size)
```
adracus, line 77: This shouldn't be checked at creation time but rather be done during a reconciliation.

ori/volume/server/volume_create.go
```diff
@@ -16,261 +16,143 @@ package server
...
+	labels[VolumeClassLabel] = volume.Spec.Class
 
-		pvc.Spec.DataSourceRef = &corev1.TypedObjectReference{
-			APIGroup: pointer.String("snapshot.storage.k8s.io"),
-			Kind:     "VolumeSnapshot",
-			Name:     imageName,
+	var snapshotRef *string
+	var snapshot *api.Snapshot
+	if image := volume.Spec.Image; image != "" {
+		img, err := s.registry.Resolve(ctx, image)
```
adracus, line 63: The image shouldn't be resolved at creation time. If an image e.g. becomes available later, the volume should stay in state = Pending and then be populated once available. Your process of creating the snapshot / using it will have to be adjusted to fit this case.

ori/volume/server/volume.go
```diff
@@ -16,235 +16,127 @@ package server
...
+		Metadata: metadata,
+		Spec:     spec,
+		Status: &ori.VolumeStatus{
+			State:  state,
+			Access: access,
+		},
+	}, nil
+}
+
+func wwnFromImageID(id string) string {
```
adracus, line 74: With this, you're making implicit assumptions on the generated name of the image below. What if the name length becomes shorter with a new release? Panic?
The wwn generation has to work regardless of the image name.

ori/volume/server/volume_delete.go
```diff
@@ -18,27 +18,22 @@ import (
...
-		return nil, err
-	}
+	log := s.loggerFrom(ctx)
 
 	log.V(1).Info("Deleting volume")
-	if err := s.client.Delete(ctx, onmetalVolume.Pvc); err != nil {
-		if !apierrors.IsNotFound(err) {
+	if err := s.imageStore.Delete(ctx, req.VolumeId); err != nil {
+		if !errors.Is(err, store.ErrNotFound) {
 			return nil, fmt.Errorf("error deleting onmetal volume: %w", err)
```
adracus, line 34: You are not deleting an `onmetal volume` here.

adracus (changes requested):

ori/volume/server/volume.go
```diff
@@ -16,235 +16,150 @@ package server
...
-			monitors = cluster.Monitors
-			break
-		}
-	}
-	if len(monitors) == 0 {
-		return nil, fmt.Errorf("no monitors provided for clusterID %s", s.rookPoolName)
-	}
-	return monitors, nil
+func (s *Server) generateWWN() string {
+	wwn := strings.Replace(s.idGen.Generate(), "-", "", -1)
```
adracus, line 86: Two things:
1. `IDGen` by default doesn't return `-` in its output, so the `strings.Replace` is not necessary.
2. I'd recommend using a separate instance of `IDGen` for WWN generation. This instance should then by default produce strings of length 16 which also makes the `[:16]` unnecessary

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,92 @@ package server
...
-		return nil, fmt.Errorf("failed to resolve image ref in registry: %w", err)
-	}
-
-	layers, err := img.Layers(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get layers for image: %w", err)
-	}
+func (s *Server) getCephVolumeConfig(ctx context.Context, log logr.Logger, volume *ori.Volume) (*CephVolumeConfig, error) {
+	log.V(2).Info("Getting ceph volume config")
+	defer log.V(2).Info("Successfully got ceph volume config")
```
adracus, line 34: Remove the `defer` - if you're returning an `error`, the log would still be printed which doesn't make much sense.

ori/volume/server/volume_create.go
```diff
@@ -17,260 +17,92 @@ package server
...
-		return fmt.Errorf("error patching pvc as created: %w", err)
+func (s *Server) createImage(ctx context.Context, log logr.Logger, cfg *CephVolumeConfig) (retErr error) {
+	image, err := s.imageStore.Create(ctx, cfg.image)
+	if err != nil {
+		return fmt.Errorf("failed to create image: %w", err)
 	}
 
-	// Reset cleaner since everything from now on operates on a consistent volume
-	c.Reset()
+	cfg.image = image
```
adracus, line 82: Don't mutate the `cfg.image` here but instead return an `AggregatedImage` (if that type doesn't exist, create it).
Idea behind is the following: For creating, you first calculate all necessary artefacts (which results in a `CephVolumeConfig`). This is the parameter for the `createImage` call which produces an `AggregatedImage` (internal image with additional info required for ori). This way, there's always a clear separation between artefacts required for creation and the result of creation / getting existing images.

ori/volume/server/volume_list.go
```diff
@@ -148,9 +53,29 @@ func (s *Server) filterVolumes(volumes []*ori.Volume, filter *ori.VolumeFilter)
...
 
+func (s *Server) listVolumes(ctx context.Context, log logr.Logger) ([]*ori.Volume, error) {
+	cephImages, err := s.imageStore.List(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error listing volumes: %w", err)
+	}
+
+	var res []*ori.Volume
+	for _, cephImage := range cephImages {
+		oriVolume, err := s.convertImageToOriVolume(ctx, log, cephImage)
```
adracus, line 64: If `imageStore.List` returns a volume that is not managed by the `Server`, the whole `List` operation errors. `listVolumes` should filter out volumes that are not created / managed via `ori` (for reference, e.g. the `machinebroker` filters by managed and created `Machine`s in `onmetal-api`).

pkg/controllers/snapshot_controller.go
```diff
@@ -0,0 +1,394 @@
...
+	buffer := make([]byte, r.populatorBufferSize)
+	_, err := io.CopyBuffer(dst, rater, buffer)
+	if err != nil {
+		_ = dst.Close()
+		return fmt.Errorf("failed to populate image: %w", err)
+	}
+	log.Info("Successfully populated image")
+
+	return nil
+}
```
adracus, line 394: `dst` is not closed on success. Also, `populateImage` should not call `dst.Close` in the first place, as it doesn't actually manage the `io.WriteCloser` since the calling method is the entity opening / having to close the image.
In general, for operations involving `io.Closer`, either you return the `io.Closer` or the function opening it should be the one closing it, which means `Close()` should be called in `reconcileSnapshot`.

pkg/controllers/image_controller.go
```diff
@@ -0,0 +1,472 @@
...
+
+	return strings.TrimPrefix(r.client, "client."), response.Key, nil
+}
+
+func (r *ImageReconciler) reconcileSnapshot(ctx context.Context, log logr.Logger, img *api.Image) error {
+	if img.Spec.Image == "" || img.Spec.SnapshotRef != nil {
+		return nil
+	}
+
+	resolvedImg, err := r.registry.Resolve(ctx, img.Spec.Image)
```
adracus, line 269: Why is the image still resolved if a `snapshotRef` is already present? That shouldn't happen.

pkg/controllers/image_controller.go
```diff
@@ -0,0 +1,472 @@
...
+		case errors.Is(err, store.ErrNotFound):
+			snap, err = r.snapshots.Create(ctx, &api.Snapshot{
+				Metadata: api.Metadata{
+					ID: snapshotDigest,
+					Labels: map[string]string{
+						imageDigestLabel: snapshotDigest,
+					},
+				},
+				Source: api.SnapshotSource{
+					OnmetalImage: img.Spec.Image,
```
adracus, line 291: The source should point to the resolved snapshot, not again to the image leading it to be potentially resolved to another digest.

pkg/controllers/image_controller.go
```diff
@@ -0,0 +1,472 @@
...
+	img, err := librbd.OpenImage(ioCtx, ImageIDToRBDID(image.ID), librbd.NoSnapshot)
+	if err != nil {
+		return fmt.Errorf("failed to open rbd image: %w", err)
+	}
+
+	for limit, value := range image.Spec.Limits {
+		if err := img.SetMetadata(fmt.Sprintf("%s%s", LimitMetadataPrefix, limit), strconv.FormatInt(value, 10)); err != nil {
+			return fmt.Errorf("failed to set limit (%s): %w", limit, err)
+		}
+		log.V(3).Info("Set image limit", "limit", limit, "value", value)
```
adracus, line 410: `img` is not closed in case a `SetMetadata` call fails.

pkg/controllers/image_controller.go
```diff
@@ -0,0 +1,472 @@
...
+		return false, fmt.Errorf("failed to clone rbd image: %w", err)
+	}
+
+	img, err := librbd.OpenImage(ioCtx, ImageIDToRBDID(image.ID), librbd.NoSnapshot)
+	if err != nil {
+		return false, fmt.Errorf("failed to open rbd image: %w", err)
+	}
+
+	if err := img.Resize(uint64(image.Spec.Size)); err != nil {
+		return false, fmt.Errorf("failed to resize rbd image: %w", err)
```
adracus, line 462: `img` is not closed in case `Resize` fails.

ori/volume/server/volume_create.go
adracus: With this we still have the problem of the size not matching with the given one. What's about the discussion we had that at least the spec reports the specified size and only the reconciler doing the 2MB rounding?

adracus (changes requested):

pkg/controllers/image_controller.go
lukasfrank: see 2-3 lines above, this case will be catched there:
```
img.Spec.Image == "" || img.Spec.SnapshotRef != nil
```

ori/volume/server/volume_list.go
lukasfrank: `imageStore.List(ctx)` will only return the content of the `omap` hence it is still possible to create `ceph images` directly without interfering here. Since a `VolumePool` is corresponding to one `ceph pool` two instances also will not conflict (since omaps are not shared across pools).

adracus: But what if I create an `api.Image` using the internal API only (i.e. `imageStore.Create`) and this `api.Image`. This `api.Image` was not created via `ori` and should not appear in the result set of `listVolumes` but currently it would. Even worse, as far as I can see it'd cause an error since using the internal API, I'm not obliged to specify the `volumeClassLabel`, resulting in an error upon calling `convertImageToOriVolume`.

adracus (approved):

.github/workflows/publish-docker.yml
```diff
@@ -26,8 +24,6 @@ jobs:
         image:
           - name: cephlet
             target: manager
-          - name: populator
-            target: populator
           - name: ori-volume
```
afritzler, line 27: Is this the correct image name?

.github/workflows/golangci-lint.yml
```diff
@@ -12,9 +12,10 @@ jobs:
     steps:
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.19'
+          go-version: '1.20'
+      - run: sudo apt update  && sudo apt install -y libcephfs-dev librbd-dev librados-dev
       - uses: actions/checkout@v3
```
afritzler, line 17: Move the checkout step before setup-go

.github/workflows/golangci-lint.yml
```diff
@@ -12,9 +12,10 @@ jobs:
     steps:
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.19'
+          go-version: '1.20'
```
afritzler, line 15: Use here `go-version-file: 'go.mod'` instead of the version.

.github/workflows/test.yml
```diff
@@ -15,5 +15,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.19'
+          go-version: '1.20'
```
afritzler, line 18: Use `go-version-file: 'go.mod'` instead.

pkg/rook/config.go
```diff
@@ -15,79 +15,38 @@
...
-	StorageClassReclaimPolicyDefaultValue        = "Delete"
-	StorageClassVolumeBindingModeDefaultValue    = "Immediate"
-	CSIDriverNameDefaultValue                    = "rook-ceph.rbd.csi.ceph.com"
-	NamespaceDefaultValue                        = "rook-ceph"
-	EnableRBDStatsDefaultValue                   = false
-	BurstFactorDefaultValue                      = 10
-	BurstDurationInSecondsDefaultValue           = 15
-
-	BucketProvisionerDefaultValue = "rook-ceph.ceph.rook.io/bucket"
+	ClusterIdDefaultValue                 = "rook-ceph"
```
afritzler, line 18: Is there a reason why we will have the `rook` constants here?

afritzler (changes requested):

pkg/rook/config.go
lukasfrank: This stuff is needed for the `Bucket` reconciler and will be removed in the next refactoring.

.github/workflows/publish-docker.yml
lukasfrank: the populator has been removed and is now covered by the `Snapshot` controller

afritzler (approved):

PR #280, opened 2023-06-16
cephlet-volume e2e tests

lukasfrank:
# Proposed Changes

-
-
-

Fixes #


e2e/Makefile
```diff
@@ -2,7 +2,8 @@
 
 .PHONY: create-env
 create-env:  ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	curl -sfL https://get.k3s.io | sh -s - --docker 
+	curl -sfL https://get.k3s.io | sh -s - --docker
```
lukasfrank, line 5: @pradumnapandit I think that we need to stick to `minikube` since we need to attach disks to a vm in order to run ceph. Please check out this documentation: https://rook.io/docs/rook/v1.11/Contributing/development-environment/?h=minikube#minikube

e2e/volume_test.go
```diff
@@ -0,0 +1,101 @@
...
+				VolumePoolRef:  &corev1.LocalObjectReference{Name: "ceph"},
+				Resources: corev1alpha1.ResourceList{
+					corev1alpha1.ResourceStorage: resource.MustParse(volumeSize),
+				},
+			Image: "ghcr.io/onmetal/onmetal-image/gardenlinux:rootfs-dev-20230223",
+			},
+		}
+		Expect(k8sClient.Create(ctx, vol)).To(Succeed())
+		fmt.Println("Here the Volume is getting created############", vol.Name)
+	})
```
lukasfrank, line 70: The `onmetal-api` is already tested in a different place. Rather create a `Volume` and check in the ceph cluster for instance if a `ceph image` got created, the image has the correct meta data and so on.

e2e/volume_test.go
```diff
@@ -0,0 +1,101 @@
...
+		volume := &storagev1alpha1.Volume{}
+		ns := types.NamespacedName{Namespace: "rook-ceph", Name: "tsi"}
+		Expect(k8sClient.Get(ctx, ns, volume)).To(Succeed())
+		fmt.Println("Here the Volume is getting listed##############", volume.Name)
+
+		// Todo use matcher
+		//Expect(volume.Name).To(Succeed())
+
+		//Expect(k8sClient.List(ctx, volumeList, client.InNamespace("rook-ceph"))).To(Succeed())
+	})
```
lukasfrank, line 82: same here: this test do not cover the functionality of the `cephlet`

e2e/Makefile
pradumnapandit: Okay, will try to cover with`minikube`.

e2e/volume_test.go
pradumnapandit: Okay.. We will add that part as well.

e2e/volume_test.go
pradumnapandit: Will Add.

e2e/Makefile
pradumnapandit: Added the ENV creation using minikube.

PR #348, opened 2023-09-18
Cephlet migration

pradumnapandit:
# Migration script details from old cephlet to new cephlet (Updates in OMAP data)
- Updated the **OMAP data** with migration script 
- Created **Dockerfile** for migration script
- Created deployment file with default **dry_run** flag enabled


lukasfrank:
@FlorinPeter Do we still need this PR? I would propose to move this to an internal repo/wiki.

pradumnapandit:
I guess, we can merge as well, as folder structure is not impacting to main.

lukasfrank:
> I guess, we can merge as well, as folder structure is not impacting to main.

I think it's not a good idea to pollute the project. However, we can keep the PR open until the migration on your side is done. Our latest and only release (https://github.com/onmetal/cephlet/releases/tag/v0.1.0) uses already the ORI without the k8s CI.

pradumnapandit:
Yes. However, OMAP changes are major changes in migration and its script, so related changes would keep a track in the specific folder, which I thought of.

lukasfrank:
As clarified with @dnlvanco, closing this PR

PR #349, opened 2023-09-19
Add Integration tests for `CreateVolume` 

### LINKED ISSUE #339: CreateVolume Method Testing
# Summary
Test CreateVolume Method

## Description
- Create test cases to validate the functionality of the `CreateVolume` method.
- Utilize the `go-ceph` library to verify that volumes are properly created in the Ceph cluster with different configurations.
- Check for error conditions and boundary cases.

kasabe28:
# Proposed Changes

- Add integration test for create volume 
- Add integration test for create encrypted volume

Fixes #339


integration/integration_suite_test.go
```diff
@@ -78,13 +79,14 @@ var _ = BeforeSuite(func() {
...
-			KeyringFile:          os.Getenv("CEPH_KEYRING_FILENAME"),
-			Pool:                 os.Getenv("CEPH_POOLNAME"),
-			Client:               os.Getenv("CEPH_CLIENTNAME"),
-			KeyEncryptionKeyPath: keyEncryptionKeyFile.Name(),
+			ConnectTimeout:         10 * time.Second,
+			Monitors:               os.Getenv("CEPH_MONITORS"),
+			User:                   os.Getenv("CEPH_USERNAME"),
+			KeyringFile:            os.Getenv("CEPH_KEYRING_FILENAME"),
+			Pool:                   os.Getenv("CEPH_POOLNAME"),
+			Client:                 os.Getenv("CEPH_CLIENTNAME"),
```
afritzler, line 87: As an idea: can we set those values as global variables on top as they are also used in the subsequent tests like create volume test? That way we ensure that we only use the `Getenv` only once upon suite creation.

integration/volume_create_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+				HaveKeyWithValue(api.BPSBurstDurationLimit, int64(15)),
+				HaveKeyWithValue(api.IOPSlLimit, int64(100)),
+			)),
+			HaveField("Spec.SnapshotRef", BeNil()),
+			HaveField("Spec.Encryption.Type", api.EncryptionTypeUnencrypted),
+			HaveField("Spec.Encryption.EncryptedPassphrase", BeEmpty()),
+			HaveField("Status.State", Equal(api.ImageStateAvailable)),
+			HaveField("Status.Access", SatisfyAll(
+				HaveField("Monitors", os.Getenv("CEPH_MONITORS")),
+				HaveField("Handle", fmt.Sprintf("%s/%s", os.Getenv("CEPH_POOLNAME"), "img_"+createResp.Volume.Metadata.Id)),
```
afritzler, line 89: As described in the comment above: use the global variables here instead of reading the environment variables.

afritzler (changes requested):

afritzler (approved):

PR #357, opened 2023-09-25
Add integration test for `ListVolumes`

### LINKED ISSUE #338: ListVolumes Method Testing
# Summary
Test ListVolumes Method

## Description
- Create test cases to validate the functionality of the `ListVolumes` method.
- Use the `go-ceph` library to check if the listed volumes in the Ceph cluster match the expected results.
- Ensure it can list volumes correctly under various scenarios.
- Check for edge cases and error handling.

Rohit-0505:
- Add integration test for `ListVolumes` method

Fixes #338

### DIFF
--- a/integration/volume_list_integration_test.go
+++ b/integration/volume_list_integration_test.go
@@ -1,0 +1,122 @@
+// Copyright 2023 OnMetal authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


afritzler (changes requested):
How about adding also a negative test e.g. label selector which doesn't find any volumes.

afritzler (approved):

PR #375, opened 2023-10-12
Add tests for `DeleteBucket`

### LINKED ISSUE #362: Test `DeleteBucket` Method in BucketRuntime Service
# Summary
Test `DeleteBucket` Method in BucketRuntime Service

## Description
Create test cases to validate the `DeleteBucket` method in the `BucketRuntime` service. Verify that it correctly deletes the specified bucket and handles cases where the bucket does not exist or cannot be deleted due to dependencies. Test error scenarios and check if the method provides meaningful error messages.

### Instructions:
1. Install the necessary CRDs in the test environment.
2. Write test cases to confirm the functionality of the `DeleteBucket` method.
3. Ensure that the tests only patch the `rook` resources for testing purposes.
4. Test scenarios should include successful deletion, deletion of non-existing buckets, and deletion failures due to dependencies or incorrect parameters.
5. Validate error responses for different failure scenarios.

kasabe28:
# Proposed Changes

- Add tests for `DeleteBucket` function Fixes #362

### DIFF
--- a/ori/bucket/server/bucket_delete_test.go
+++ b/ori/bucket/server/bucket_delete_test.go
@@ -1,0 +1,136 @@
+// Copyright 2023 OnMetal authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server_test
+
+import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,6 +63,7 @@ vet: ## Run go vet against code.
 .PHONY: test
 test: manifests generate fmt vet envtest checklicense ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go run github.com/onsi/ginkgo/v2/ginkgo --label-filter="!integration" -coverprofile cover.out ./...
+	go mod tidy
```
lukasfrank, line 66: do we need this command?

lukasfrank (changes requested):

Makefile
kasabe28: Yes. As it cleans test related dependencies from go.sum file after tests are run. This is added in other repositories as well like `gardener-extension-provider-onmetal`.

lukasfrank: I think we should remove it since it seems to be a temporary fix. The pipeline also checks if the go.mod is up to-date.

kasabe28: done

lukasfrank (approved):

PR #391, opened 2023-11-02
Remove BucketClass API dependency

### LINKED ISSUE #390: Remove `BucketClass` API dependency
# Summary

Currently the `Bucket` server relies on the `BucketClass` API to determine which `BucketClass` are supported. This should change to reflect the same behaviour as in the `Volume` server. Here we are using a configuration file to announce the supported `VolumeClasses`.

## Tasks
- Implement a config based `BucketClass` announcement to the `Bucket` server similar to the `Volume` setup
- Adjust the test suite setup and the affected test files to reflect those changes

ushabelgur:
# Proposed Changes

-  Remove BucketClass API dependency
-  Configure BucketClass announcement to the Bucket server similar to the Volume setup
-  Fix test suite setup and the affected test files to reflect the changes

Fixes #390

### DIFF
--- a/ori/bucket/server/suite_test.go
+++ b/ori/bucket/server/suite_test.go
@@ -16,6 +16,7 @@ package server_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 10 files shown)


pkg/bcr/registry.go
```diff
@@ -0,0 +1,76 @@
...
+	var classList []ori.BucketClass
+	if err := yaml.NewYAMLOrJSONDecoder(reader, 4096).Decode(&classList); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal bucket classes: %w", err)
+	}
+
+	return classList, nil
+}
+
+func LoadBucketClassesFile(filename string) ([]ori.BucketClass, error) {
+	file, err := os.Open(filename)
```
afritzler, line 37: You need to close the file handle at some point by calling `file.Close()`. Ideally this has to be done via a `defer` call.

afritzler (changes requested):

ori/bucket/server/suite_test.go
```diff
@@ -87,21 +83,39 @@ var _ = BeforeSuite(func() {
...
+
+	bucketClassesData, err := json.Marshal(bucketClasses)
+	Expect(err).NotTo(HaveOccurred())
 
-	cfg, err = envtestutils.StartWithExtensions(testEnv, testEnvExt)
+	bucketClassesFile, err := os.CreateTemp(GinkgoT().TempDir(), "bucketClasses")
+	Expect(err).NotTo(HaveOccurred())
+	defer func() {
+		_ = bucketClassesFile.Close()
+	}()
```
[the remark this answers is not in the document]
ushabelgur: Done

pkg/bcr/registry.go
ushabelgur: Done

afritzler (approved):

PR #490, opened 2024-03-20
Refactor project structure

afritzler:
# Proposed Changes

Refactore the current project structure to reflect the current conventions in Go. Create top-level packages for `api`, `cmd` and keep the implementation details in the `internal` package.

Also bump golang version to 1.22 and update the `golangci-lint` to v1.57.0.


internal/snapshotstrategy/strategy.go
```diff
@@ -1,12 +1,12 @@
 // SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and IronCore contributors
 // SPDX-License-Identifier: Apache-2.0
 
-package utils
+package snapshotstrategy
```
lukasfrank, line 4: I think it's not 100% correct, to have both strategies in the same package. Should we maybe rename it to `strategy`?

internal/bucketserver/bucket.go
```diff
@@ -1,13 +1,13 @@
 // SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and IronCore contributors
 // SPDX-License-Identifier: Apache-2.0
 
-package server
+package bucketserver
```
lukasfrank, line 4: Wdyt about putting the `bucketserver` and `volumeserver` in a `server` package to keep the package name concise?

lukasfrank (approved):

PR #540, opened 2024-05-24
Added `ResourceVersion`

### LINKED ISSUE #461: Race Condition in store
**Describe the bug**
The `omap` store has a race condition and hence metadata can be lost.  
Since the store can be accessed in parallel, data can be fetched, updated and overwrite changes from a concurrent operation.

pradumnapandit:
Fixes https://github.com/ironcore-dev/ceph-provider/issues/461

### DIFF
--- a/internal/omap/omap.go
+++ b/internal/omap/omap.go
@@ -246,7 +246,6 @@ func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
 	}
 	obj.IncrementResourceVersion()
 
-	//Todo: update version
 	obj, err = s.set(ioCtx, obj)
 	if err != nil {
 		return utils.Zero[E](), err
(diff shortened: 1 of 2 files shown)


internal/omap/omap.go
```diff
@@ -237,6 +241,11 @@ func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
 		return obj, nil
 	}
 
+	if oldObj.GetResourceVersion() != obj.GetResourceVersion() {
+		return utils.Zero[E](), fmt.Errorf("failed to update object: %w", ErrResourceVersionNotLatest)
+	}
+	obj.IncrementResourceVersion()
+
 	//Todo: update version
```
lukasfrank, line 249: Can you please remove this todo?

lukasfrank (changes requested):

internal/omap/omap.go
pradumnapandit: Removed Todo.

lukasfrank (approved):

PR #570, opened 2024-08-08
Implement ListEvents Runtime Interface to Support Cross-Cluster Events

### LINKED ISSUE #563: Implement `ListEvents` Runtime Interface to Support Cross-Cluster Events
# Summary

This proposal aims to implement the Runtime interface to support cross-cluster events, enhancing observability and debug-ability (see also https://github.com/ironcore-dev/ironcore/issues/1070).

## Basic example

Events of the `ceph-provider` can be stored in-memory until they are fetched by the `poollet`. To avoid exceeding memory limits, events should be garbage collected after a defined TTL and the event store should evict the oldest events in case of reaching the max capacity.

ushabelgur:
# Proposed Changes

- Add ListEvents implementation to volume provider
- Add event recorder store implementation 

Fixes #563


internal/controllers/image_controller.go
```diff
@@ -235,7 +242,7 @@ type fetchAuthResponse struct {
 	Key string `json:"key"`
 }
 
-func (r *ImageReconciler) fetchAuth(ctx context.Context, log logr.Logger) (string, string, error) {
```
lukasfrank, line 238: Can we please keep the `ctx`? Currently, a lot of the `ceph` commands do not support a `ctx`, however libraries tend to add a `ctx` from time to time. If we run into issues we can also implement our cancelation logic.

lukasfrank (changes requested):

internal/controllers/image_controller.go
ushabelgur: reverted back changes

lukasfrank:
Since `reconcileSnapshot` influences the processing time of the `image` provisioning it would be great to add more context by writing events there.

ushabelgur:
> Since `reconcileSnapshot` influences the processing time of the `image` provisioning it would be great to add more context by writing events there.

Volume metadata is attached only with image, its not being propagated to snapshots currently. If i generate events for snapshot with this volume metadata information, it will not be useful in propagating event. Will try to check if i can populate this info in snapshot.

lukasfrank:
> > Since `reconcileSnapshot` influences the processing time of the `image` provisioning it would be great to add more context by writing events there.
> 
> Volume metadata is attached only with image, its not being propagated to snapshots currently. If i generate events for snapshot without this volume metadata information, it will not be useful in propagating event. Will try to check if i can populate this info in snapshot.

Can't we say here, that the snapshot was not present and therefore we gonna create it. I think this is still useful information and we also have the image.

lukasfrank (approved):

PR #603, opened 2024-10-10
Bump `k8s.io/*` deps to v0.31.1

lukasfrank:
# Proposed Changes

- Bump k8s.io/* deps to v0.31.1
- Bump golang to 1.23.0
- Bump golangci-lint 1.61


.github/workflows/golangci-lint.yml
```diff
@@ -18,5 +18,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.57.2
-          skip-pkg-cache: true
+          version: v1.61.0
```
afritzler, line 21: You can just use `v1.61` here so we don't need to worry about the patch versions in the GH workflows.

afritzler (changes requested):
Some minor thing in the linter workflow. Apart from that the PR looks good.

afritzler (approved):

PR #737, opened 2025-06-23
Enhance error statements to return proper GRPC error codes

### LINKED ISSUE #733: Return proper GRPC error code for IRI calls
We need to check if we are handling properly grpc error. We should always set proper grpc error before return error response. Ideally we will need to prepare some mapping logic between internal and grpc errors.

Basic example

2025-06-13T06:27:32.336749927Z  ERROR   iri-server./machine.v1alpha1.MachineRuntime/CreateMachine       Unknown error handling request  {"error": "unable to get libvirt machine config: cannot allocate resources: failed to allocate resource memory: not enough available resources"}

We should return proper grpc error code and not the Unknown error

Refer to this commit : 
https://github.com/ironcore-dev/ironcore/pull/1309

Rohit-0505:
# Proposed changes
- Update all the IRI methods to ensure they return the proper GRPC error codes in the error statements

Fixes #733

### DIFF
--- a/internal/bucketserver/bucket.go
+++ b/internal/bucketserver/bucket.go
@@ -8,10 +8,9 @@ import (
 	"fmt"
 
 	"github.com/ironcore-dev/ceph-provider/api"
+	"github.com/ironcore-dev/ceph-provider/internal/utils"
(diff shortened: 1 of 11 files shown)


lukasfrank (approved):
LGTM

PR #811, opened 2025-11-20
Flatten cloned images of snapshot during snapshot deletion

### LINKED ISSUE #809: Unable to delete `VolumeSnapshot` and source `Volume` used by restored `Volume`
**Describe the bug**
Unable to delete `VolumeSnapshot` if used as a `dataSource` to restore `Volume`. Also source `Volume` cannot be deleted if it has `VolumeSnapshot`s. To delete `Volume` user needs to delete all `VolumeSnapshot`s first and then `Volume` can be deleted.

**To Reproduce**
1. Create `Volume`
2. Create `VolumeSnapshot`
3. Restore `Volume` using `VolumeSnapshot` as `dataSource`
4. Delete `VolumeSnapshot`
5. Delete source `Volume` from step 1

**Expected behavior**
A `VolumeSnapshot` and source `Volume` should be deleted successfully

**Additional context**
`Ceph` implementation does not allow to delete `Volume` if it has `VolumeSnapshot`s.

kasabe28:
# Proposed Changes

- Flatten cloned images of snapshot while deleting snapshot
- Delete volume only if it has no snapshot.
- Delete volume os-image only if it is not referenced by any other volumes
- Improve error handling in `CreateVolume()`

Fixes #809

### DIFF
--- a/internal/controllers/snapshot_controller.go
+++ b/internal/controllers/snapshot_controller.go
@@ -160,7 +160,7 @@ const (
 	SnapshotFinalizer = "snapshot"
 )
 
-func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, snapshotID string, img *librbd.Image) error {
+func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, ioCtx *rados.IOContext, snapshotID string, img *librbd.Image) error {
 	log.V(2).Info("Remove snapshot")
 
 	pools, imgs, err := img.ListChildren()
@@ -169,8 +169,22 @@ func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, snapshotID string, 
 	}
 	log.V(2).Info("Snapshot references", "pools", len(pools), "rbd-images", len(imgs))
 
-	if len(pools) != 0 || len(imgs) != 0 {
-		return fmt.Errorf("unable to delete snapshot: still in use")
+	for _, clonedImgName := range imgs {
+		log.V(2).Info("Open cloned image to flatten", "ClonedImageID", clonedImgName)
+		img, err := librbd.OpenImage(ioCtx, clonedImgName, librbd.NoSnapshot)
+		if err != nil {
+			if !errors.Is(err, librbd.ErrNotFound) {
+				return fmt.Errorf("failed to open cloned image: %w", err)
+			}
+			continue
+		}
+		defer closeImage(log, img)
+
+		log.V(2).Info("Flatten cloned image", "ClonedImageID", clonedImgName)
+		if err := img.Flatten(); err != nil {
+			return fmt.Errorf("failed to flatten cloned image: %w", err)
+		}
+		log.V(2).Info("Flattened cloned image", "ClonedImageID", clonedImgName)
 	}
 
 	snapshot := img.GetSnapshot(snapshotID)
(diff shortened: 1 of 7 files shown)


internal/controllers/snapshot_controller.go
```diff
@@ -169,8 +169,22 @@ func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, snapshotID string,
...
+		img, err := librbd.OpenImage(ioCtx, clonedImgName, librbd.NoSnapshot)
+		if err != nil {
+			if !errors.Is(err, librbd.ErrNotFound) {
+				return fmt.Errorf("failed to open cloned image: %w", err)
+			}
+			continue
+		}
+		defer closeImage(log, img)
+
+		log.V(2).Info("Flatten cloned image", "ClonedImageID", clonedImgName)
```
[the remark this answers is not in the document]
kasabe28: We don't have snapshot event recorder. Maybe we can add event recorder and record all snapshot events in a separate PR.

internal/controllers/snapshot_controller.go
```diff
@@ -169,8 +169,22 @@ func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, snapshotID string,
...
+		if err := img.Flatten(); err != nil {
```
balpert89: As flattening an image adds considerable time consumption we may want to limit parallel flattening operations.

internal/controllers/snapshot_controller.go
```diff
@@ -169,8 +169,22 @@ func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, snapshotID string,
 	}
 	log.V(2).Info("Snapshot references", "pools", len(pools), "rbd-images", len(imgs))
 
-	if len(pools) != 0 || len(imgs) != 0 {
-		return fmt.Errorf("unable to delete snapshot: still in use")
+	for _, clonedImgName := range imgs {
+		log.V(2).Info("Open cloned image to flatten", "ClonedImageID", clonedImgName)
+		img, err := librbd.OpenImage(ioCtx, clonedImgName, librbd.NoSnapshot)
```
sujeet01: `img` in the loop here shadows the above function parameter `img` variable. Please use  different variable name in the loop.

```suggestion
		clonedImg, err := librbd.OpenImage(ioCtx, clonedImgName, librbd.NoSnapshot) 
```

internal/controllers/snapshot_controller.go
```diff
@@ -169,8 +169,22 @@ func (r *SnapshotReconciler) removeSnapshot(log logr.Logger, snapshotID string,
...
+		if err != nil {
+			if !errors.Is(err, librbd.ErrNotFound) {
+				return fmt.Errorf("failed to open cloned image: %w", err)
+			}
+			continue
+		}
+		defer closeImage(log, img)
```
sujeet01: We should not use `defer` in the for loop, since `defer` does not get executed until the function returns. This will keep many images open if we have multiple clones, causing resource exhaustion. 
In the for loop here, we need to explicitly close the image after use (after flattening). 

Better approach would be to extract the open/flatten logic into a helper function where `defer` can be used efficiently.
```suggestion
```

kasabe28:
As now we have another PR for source `Volume` and `VolumeSnapshot` deletion, which has changes from this PR also. I have addressed all comments in another PR -> https://github.com/ironcore-dev/ceph-provider/pull/813.

PR #813, opened 2025-11-24
Clone snapshots and flatten rbd images before deletion

### LINKED ISSUE #809: Unable to delete `VolumeSnapshot` and source `Volume` used by restored `Volume`

kasabe28:
# Proposed Changes

- Flatten cloned images of snapshot during snapshot deletion
- Clone snapshots into images and flatten snapshot children if it has before source volume deletion
- Delete volume os-image only if it is not referenced by any other volumes
- Improve error handling in `CreateVolume()`
- Introduce helper functions - `openImage()`, `closeImage()`, `flattenImage()`, `createSnapshot()`, `removeSnapshot()`, `flattenChildImages(), isSnapshotExist()` and use it wherever it can be.

Fixes #809 

## Summary by CodeRabbit

* **Bug Fixes**
  * Deletion flows now clone snapshot sources, flatten child images, and handle finalizers to avoid orphaned data and improve error handling.

* **New Features**
  * Snapshot-based cloning creates managed images that preserve metadata and labels; cloning steps are more traceable via enhanced logs.

* **Improvements**
  * Volume/image creation derives size and optional encryption from snapshot or image sources with safer nil-handling and clearer logging.

* **Chores**
  * Image encryption spec in the public API is now nullable (optional).

* **Tests**
  * Integration tests updated to expect nil encryption for unencrypted volumes and enable more verbose test logging.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/internal/controllers/common.go
+++ b/internal/controllers/common.go
@@ -43,7 +43,7 @@ func getSnapshotSourceDetails(snapshot *providerapi.Snapshot) (parentName string
 }
 
 func closeImage(log logr.Logger, img *librbd.Image) {
-	if closeErr := img.Close(); closeErr != nil && closeErr != librbd.ErrImageNotOpen {
+	if closeErr := img.Close(); closeErr != nil && errors.Is(closeErr, librbd.ErrImageNotOpen) {
 		log.Error(closeErr, "failed to close image")
 	}
 }
@@ -137,7 +137,7 @@ func flattenChildImages(log logr.Logger, conn *rados.Conn, img *librbd.Image) er
 	return nil
 }
 
-func isSnapshotExist(log logr.Logger, ioCtx *rados.IOContext, imageName string, snapshotName string) bool {
+func snapshotExists(log logr.Logger, ioCtx *rados.IOContext, imageName string, snapshotName string) bool {
 	img, err := openImage(ioCtx, imageName)
 	if err != nil {
 		return false
--- a/internal/controllers/common.go
+++ b/internal/controllers/common.go
@@ -59,8 +59,15 @@ func openImage(ioCtx *rados.IOContext, imageName string) (*librbd.Image, error) 
 	return img, nil
 }
 
-func flattenImage(log logr.Logger, ioCtx *rados.IOContext, imageName string) error {
+func flattenImage(log logr.Logger, conn *rados.Conn, pool string, imageName string) error {
 	log.V(2).Info("Flatten cloned image", "clonedImageId", imageName)
+
+	ioCtx, err := conn.OpenIOContext(pool)
+	if err != nil {
+		return fmt.Errorf("unable to open io context: %w", err)
+	}
+	defer ioCtx.Destroy()
+
 	img, err := openImage(ioCtx, imageName)
 	if err != nil {
 		return err
@@ -85,7 +92,7 @@ func createSnapshot(log logr.Logger, ioCtx *rados.IOContext, snapshotName string
 	if err != nil {
 		return fmt.Errorf("unable to create snapshot: %w", err)
 	}
-	log.Info("snapshot created")
+	log.Info("Snapshot created")
 
 	if err := imgSnap.Protect(); err != nil {
 		return fmt.Errorf("unable to protect snapshot: %w", err)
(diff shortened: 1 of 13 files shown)


gonzolino (commented):
Review not 100% done yet, but here are some remarks I found so far.

internal/controllers/image_controller.go
```diff
@@ -228,49 +228,191 @@ func (r *ImageReconciler) deleteImage(ctx context.Context, log logr.Logger, ioCt
...
+
+		log.V(2).Info("Create snapshot clone", "snapshotId", snapName)
+		if err := r.cloneImageFromSnapshot(ctx, log, ioCtx, snapName, image); err != nil {
+			return fmt.Errorf("failed to create snapshot clone: %w", err)
+		}
+		log.V(2).Info("Created snapshot clone", "snapshotId", snapName)
+
+		if err := r.deleteImageSnapshot(ctx, log, ioCtx, snapName, snap); err != nil {
+			return fmt.Errorf("failed to delete image snapshot: %w", err)
+		}
```
gonzolino: In this section, any existing snapshots will be cloned, the cloned images will be flattened and then the snapshots deleted. Before (on line 278), there is a call to `flattenChildImagesIfAny`, which flattens all child images.

I think the flow would be clearer if all snapshots would be cloned, then all child images would be flattened, and only then snapshots would be deleted.

internal/controllers/image_controller.go
```diff
@@ -228,49 +228,191 @@ func (r *ImageReconciler) deleteImage(ctx context.Context, log logr.Logger, ioCt
...
+	}
+	return nil
+}
+
+func (r *ImageReconciler) deleteImageSnapshot(ctx context.Context, log logr.Logger, ioCtx *rados.IOContext, snapName string, snap *librbd.Snapshot) error {
+	// delete the cloned image if next steps fails
+	deleteClone := true
+	defer func() {
+		if deleteClone {
+			if err := r.images.Delete(ctx, snapName); err != nil {
```
gonzolino: I don't think it's a good idea to have cleanup code here in this function for an image that gets created in a different function. If we ever refactor the code it might happen that we change the order we call functions and it might mess with this cleanup code.

Also, instead of deleting the cloned image in case of a failure, maybe we can keep track of the cloned image and reuse it when we retry the deletion.

internal/controllers/snapshot_controller.go
```diff
@@ -218,18 +185,38 @@ func (r *SnapshotReconciler) deleteSnapshot(ctx context.Context, log logr.Logger
 	}
 	defer closeImage(log, img)
 
-	if err = r.removeSnapshot(log, snapshotID, img); err != nil {
+	if err := flattenChildImagesIfAny(log, ioCtx, img); err != nil {
```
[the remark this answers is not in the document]
kasabe28: Fixed it by opening rbd image along with snapshot name, so that it will list children originated only from that snapshot.

internal/controllers/image_controller.go
kasabe28: Flow is now changed as mentioned above.

internal/controllers/common.go
```diff
@@ -41,7 +43,109 @@ func getSnapshotSourceDetails(snapshot *providerapi.Snapshot) (parentName string
...
+	if closeErr := img.Close(); closeErr != nil && errors.Is(closeErr, librbd.ErrImageNotOpen) {
 		log.Error(closeErr, "failed to close image")
 	}
 }
+
+func openImage(ioCtx *rados.IOContext, imageName string) (*librbd.Image, error) {
+	img, err := librbd.OpenImage(ioCtx, imageName, librbd.NoSnapshot)
+	if err != nil {
+		if !errors.Is(err, librbd.ErrNotFound) {
+			return nil, fmt.Errorf("failed to open cloned image: %w", err)
```
gonzolino:
```suggestion
			return nil, fmt.Errorf("failed to open image: %w", err)
```

lukasfrank:
@coderabbitai review

internal/controllers/common.go
```diff
@@ -41,7 +43,95 @@ func getSnapshotSourceDetails(snapshot *providerapi.Snapshot) (parentName string
...
+
+func flattenImage(log logr.Logger, ioCtx *rados.IOContext, imageName string) error {
+	log.V(2).Info("Flatten cloned image", "ImageID", imageName)
+	img, err := openImage(ioCtx, imageName)
+	if err != nil {
+		return err
+	}
+	defer closeImage(log, img)
+
+	if err := img.Flatten(); err != nil {
```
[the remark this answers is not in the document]
gonzolino: The team decided to keep the `Flatten` operation here for now. At a later point in time we will restructure the code and move the operation out of the controller together with the image population.

internal/volumeserver/volume_create.go
```diff
@@ -24,66 +24,75 @@ func (s *Server) createImageFromVolume(ctx context.Context, log logr.Logger, vol
...
+				return nil, fmt.Errorf("failed to get snapshot source volume from store: %w", err)
 			}
 
-			encryptedPassphrase, err = s.keyEncryption.Encrypt(passphrase)
-			if err != nil {
-				return nil, fmt.Errorf("failed to encrypt passphrase: %w", err)
+			log.V(2).Info("Getting image size and encryption from snapshot source volume", "snapshotSourceVolumeID", snapshotSourceVolume.ID)
+			imageSize = snapshotSourceVolume.Spec.Size
+			if snapshotSourceVolume.Spec.Encryption != nil {
+				encryptionSpec = snapshotSourceVolume.Spec.Encryption
```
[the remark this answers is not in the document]
gonzolino: Tested, the encryption spec must be overwritten.

gonzolino (approved):

PR #837, opened 2026-02-21
Implement async worker queues for populate & flatten operations

### LINKED ISSUE #822: Move long-running ceph operations out of controllers
# Summary

Move long-running operations such as [image flattening](https://github.com/ironcore-dev/ceph-provider/blob/main/internal/controllers/common.go#L77) or [image population](https://github.com/ironcore-dev/ceph-provider/blob/main/internal/controllers/snapshot_controller.go#L412) out of the Image and Snapshot controllers.

Implement a worker pool (or similar pattern) to perform such operations asynchronously.

## Motivation

The Image and Snapshot controller currently include operations that are potentially long-running such as [image flattening](https://github.com/ironcore-dev/ceph-provider/blob/main/internal/controllers/common.go#L77) or [image population](https://github.com/ironcore-dev/ceph-provider/blob/main/internal/controllers/snapshot_controller.go#L412).

These operations may block the controllers over long periods of time, especially when operating on large images or snapshots with a lot of attached cloned images. By performing the operations asynchronously in dedicated workers, the controllers stay responsive even when a lot of long-running operations are happening in parallel.

sujeet01:
## Summary by CodeRabbit

* **New Features**
  * Added snapshot states: Populating, Flattening, Flattened.
  * Added image state: FlatteningChildren.
  * New async workers/queues for snapshot populate/flatten and image child-flattening with configurable worker-size flags.

* **Refactor**
  * Deletion and cleanup now run asynchronously with persisted intermediate snapshot/image states and queued work.

* **Bug Fixes**
  * Reconciliation short-circuits for in-progress states and treats missing referenced volume images as errors.

* **Tests**
  * Integration startup updated to exercise new worker-size options.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ironcore-dev/ceph-provider/pull/837?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Fixes #822

### DIFF
--- a/api/image.go
+++ b/api/image.go
@@ -19,6 +19,8 @@ type ImageState string
 const (
 	ImageStatePending   ImageState = "Pending"
 	ImageStateAvailable ImageState = "Available"
+	// ImageStateFlatteningChildren indicates deletion is waiting for child images to be flattened.
+	ImageStateFlatteningChildren ImageState = "FlatteningChildren"
 )
 
 type EncryptionState string
--- a/api/snapshot.go
+++ b/api/snapshot.go
@@ -18,10 +18,13 @@ type Snapshot struct {
 type SnapshotState string
 
 const (
-	SnapshotStatePending   SnapshotState = "Pending"
-	SnapshotStatePopulated SnapshotState = "Populated"
-	SnapshotStateReady     SnapshotState = "Ready"
-	SnapshotStateFailed    SnapshotState = "Failed"
+	SnapshotStatePending    SnapshotState = "Pending"
+	SnapshotStatePopulated  SnapshotState = "Populated"
+	SnapshotStatePopulating SnapshotState = "Populating"
+	SnapshotStateFlattening SnapshotState = "Flattening"
+	SnapshotStateFlattened  SnapshotState = "Flattened"
+	SnapshotStateReady      SnapshotState = "Ready"
+	SnapshotStateFailed     SnapshotState = "Failed"
 )
 
 type SnapshotStatus struct {
--- a/cmd/volumeprovider/app/app.go
+++ b/cmd/volumeprovider/app/app.go
@@ -59,7 +59,18 @@ type CephOptions struct {
 
 	VolumeEventStoreOptions eventrecorder.EventStoreOptions
 
+	// WorkerSize controls the number of concurrent workers for the fast reconcilers
+	// (ImageReconciler, SnapshotReconciler).
 	WorkerSize int
+
+	// SnapshotPopulateLongOpsWorkerSize is the number of concurrent workers used for snapshot populate long-running operations.
+	SnapshotPopulateLongOpsWorkerSize int
+
+	// SnapshotFlattenLongOpsWorkerSize is the number of concurrent workers used for snapshot flatten long-running operations.
+	SnapshotFlattenLongOpsWorkerSize int
+
+	// ImageFlattenLongOpsWorkerSize is the number of concurrent workers used for image flatten long-running operations (deletion-time flattening of children).
+	ImageFlattenLongOpsWorkerSize int
 }
 
 func (o *Options) Defaults() {
@@ -68,6 +79,9 @@ func (o *Options) Defaults() {
 	o.Ceph.BurstDurationInSeconds = 15
 	o.Ceph.PopulatorBufferSize = 5 * 1024 * 1024
 	o.Ceph.WorkerSize = 15
+	o.Ceph.SnapshotPopulateLongOpsWorkerSize = 3
+	o.Ceph.SnapshotFlattenLongOpsWorkerSize = 5
+	o.Ceph.ImageFlattenLongOpsWorkerSize = 5
 }
 
 func (o *Options) AddFlags(fs *pflag.FlagSet) {
@@ -92,7 +106,12 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.Ceph.VolumeEventStoreOptions.TTL, "volume-event-ttl", 5*time.Minute, "Time to live for volume events.")
 	fs.DurationVar(&o.Ceph.VolumeEventStoreOptions.ResyncInterval, "volume-event-resync-interval", 1*time.Minute, "Interval for resynchronizing the volume events.")
 
-	fs.IntVar(&o.Ceph.WorkerSize, "worker-size", o.Ceph.WorkerSize, "Defines the factor to calculate the burst limits.")
+	fs.IntVar(&o.Ceph.WorkerSize, "worker-size", o.Ceph.WorkerSize, "Number of concurrent workers for the fast image and snapshot reconcilers.")
+
+	fs.IntVar(&o.Ceph.SnapshotPopulateLongOpsWorkerSize, "snapshot-populate-long-ops-worker-size", o.Ceph.SnapshotPopulateLongOpsWorkerSize, "Number of concurrent workers for snapshot populate long-running operations.")
+	fs.IntVar(&o.Ceph.SnapshotFlattenLongOpsWorkerSize, "snapshot-flatten-long-ops-worker-size", o.Ceph.SnapshotFlattenLongOpsWorkerSize, "Number of concurrent workers for snapshot flatten long-running operations.")
+
+	fs.IntVar(&o.Ceph.ImageFlattenLongOpsWorkerSize, "image-flatten-long-ops-worker-size", o.Ceph.ImageFlattenLongOpsWorkerSize, "Number of concurrent workers for image flatten long-running operations (deletion-time flattening of children).")
 }
 
 func (o *Options) MarkFlagsRequired(cmd *cobra.Command) {
@@ -171,6 +190,21 @@ func Run(ctx context.Context, opts Options) error {
 	if opts.Ceph.WorkerSize <= 1 {
 		err := fmt.Errorf("invalid configuration: worker-size must be greater than 1, but got %d", opts.Ceph.WorkerSize)
 		setupLog.Error(err, "Worker size validation failed")
+		return err
+	}
+	if opts.Ceph.SnapshotPopulateLongOpsWorkerSize <= 0 {
+		err := fmt.Errorf("invalid configuration: snapshot-populate-long-ops-worker-size must be greater than 0, but got %d", opts.Ceph.SnapshotPopulateLongOpsWorkerSize)
+		setupLog.Error(err, "Snapshot populate long-ops worker size validation failed")
+		return err
+	}
+	if opts.Ceph.SnapshotFlattenLongOpsWorkerSize <= 0 {
+		err := fmt.Errorf("invalid configuration: snapshot-flatten-long-ops-worker-size must be greater than 0, but got %d", opts.Ceph.SnapshotFlattenLongOpsWorkerSize)
+		setupLog.Error(err, "Snapshot flatten long-ops worker size validation failed")
+		return err
+	}
+	if opts.Ceph.ImageFlattenLongOpsWorkerSize <= 0 {
+		err := fmt.Errorf("invalid configuration: image-flatten-long-ops-worker-size must be greater than 0, but got %d", opts.Ceph.ImageFlattenLongOpsWorkerSize)
+		setupLog.Error(err, "Image flatten long-ops worker size validation failed")
 		return err
 	}
 
@@ -256,10 +290,11 @@ func Run(ctx context.Context, opts Options) error {
 		snapshotEvents,
 		encryptor,
 		controllers.ImageReconcilerOptions{
-			Monitors:   opts.Ceph.Monitors,
-			Client:     opts.Ceph.Client,
-			Pool:       opts.Ceph.Pool,
-			WorkerSize: opts.Ceph.WorkerSize,
+			Monitors:          opts.Ceph.Monitors,
+			Client:            opts.Ceph.Client,
+			Pool:              opts.Ceph.Pool,
+			WorkerSize:        opts.Ceph.WorkerSize,
+			FlattenWorkerSize: opts.Ceph.ImageFlattenLongOpsWorkerSize,
 		},
 	)
 	if err != nil {
@@ -287,6 +322,8 @@ func Run(ctx context.Context, opts Options) error {
 			Pool:                opts.Ceph.Pool,
 			PopulatorBufferSize: opts.Ceph.PopulatorBufferSize,
 			WorkerSize:          opts.Ceph.WorkerSize,
+			PopulateWorkerSize:  opts.Ceph.SnapshotPopulateLongOpsWorkerSize,
+			FlattenWorkerSize:   opts.Ceph.SnapshotFlattenLongOpsWorkerSize,
 		},
 	)
 	if err != nil {
--- a/internal/controllers/snapshot_controller.go
+++ b/internal/controllers/snapshot_controller.go
@@ -17,7 +17,6 @@ import (
 	"github.com/go-logr/logr"
 	providerapi "github.com/ironcore-dev/ceph-provider/api"
 	"github.com/ironcore-dev/ceph-provider/internal/rater"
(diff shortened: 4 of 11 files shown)


internal/controllers/snapshot_controller.go
```diff
@@ -289,6 +288,12 @@ func (r *SnapshotReconciler) reconcileSnapshot(ctx context.Context, id string) e
 		return fmt.Errorf("failed to reconcile snapshot: %w", err)
 	}
 
+	// Return early if ironcore image snapshot is being populated asynchronously. The async populate
+	// worker will transition it to Ready after population completes.
+	if snapshot.Status.State == providerapi.SnapshotStatePopulating {
+		return nil
+	}
+
 	snapshot.Status.State = providerapi.SnapshotStateReady
```
kajolasabe: This can be removed

gonzolino (commented):
I am wondering if it wouldn't be better to have proper controllers to do asynchronous operations instead of worker pools.

This way we would have a standard reconciler with all the stuff controller-runtime offers. This controller would watch Snapshot and filter for states "Populating" and "Flattening", handle the long-running operations and bring the state back to "Ready" once finished.

What do you think?

api/snapshot.go
```diff
@@ -18,10 +18,12 @@ type Snapshot struct {
...
 
 const (
-	SnapshotStatePending   SnapshotState = "Pending"
-	SnapshotStatePopulated SnapshotState = "Populated"
-	SnapshotStateReady     SnapshotState = "Ready"
-	SnapshotStateFailed    SnapshotState = "Failed"
+	SnapshotStatePending    SnapshotState = "Pending"
+	SnapshotStatePopulated  SnapshotState = "Populated"
+	SnapshotStatePopulating SnapshotState = "Populating"
+	SnapshotStateFlattening SnapshotState = "Flattening"
```
gonzolino, line 24: The new `Populating` and `Flattening` states need to be handled in `getIriSnapshotState()` in `internal/volumeserver/snapshot.go`. Currently that switch only covers `Ready`, `Pending`, and `Failed` — any snapshot in one of these new states will cause Get/List API calls to return an error (`"unknown volume snapshot state"`). Both should map to the IRI pending state.

api/image.go
```diff
@@ -50,12 +50,20 @@ type EncryptionSpec struct {
...
 type ImageStatus struct {
-	State      ImageState      `json:"state"`
-	Encryption EncryptionState `json:"encryption"`
-	Access     *ImageAccess    `json:"access"`
-	Size       uint64          `json:"size"`
+	State         ImageState          `json:"state"`
+	Encryption    EncryptionState     `json:"encryption"`
+	Access        *ImageAccess        `json:"access"`
+	Size          uint64              `json:"size"`
+	DeletionPhase *ImageDeletionPhase `json:"deletionPhase,omitempty"`
```
gonzolino: This PR adds `SnapshotStateFlattening` directly to `SnapshotState` for snapshots, but introduces a separate `DeletionPhase` pointer field for images. This creates an inconsistency — the image's actual state becomes a cross-product of `State`, `DeletedAt`, and `DeletionPhase`. Consider adding `ImageStateFlattening` to `ImageState` instead, which would be consistent with the snapshot approach, eliminate the pointer nil-checks.

internal/controllers/snapshot_longops_controller.go
```diff
@@ -0,0 +1,432 @@
...
+		}
+	}
+	rc, _, _, err := openIroncoreImageSource(ctx, snapshot.Source.IronCoreImage, platform)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open snapshot source: %w", err)
+	}
+	return rc, nil
+}
+
+func openIroncoreImageSource(ctx context.Context, imageReference string, platform *ocispec.Platform) (io.ReadCloser, uint64, string, error) {
```
gonzolino: `openIroncoreImageSource` is defined here as a package-level function, but the same function already exists as a receiver method on `SnapshotReconciler` at `snapshot_controller.go:327`. This looks like a leftover from the refactoring — one copy should be removed and the other shared.

internal/controllers/snapshot_longops_controller.go
```diff
@@ -0,0 +1,432 @@
...
+	ioCtx, err := r.conn.OpenIOContext(r.pool)
+	if err != nil {
+		return fmt.Errorf("failed to open IO context: %w", err)
+	}
+	defer ioCtx.Destroy()
+
+	// If the Ceph snapshot already exists, population likely completed previously but status update didn't.
+	rbdImageID := SnapshotIDToRBDID(snapshot.ID)
+	exists, _, err := snapshotExistsAndProtected(log, ioCtx, rbdImageID, ImageSnapshotVersion)
+	if err == nil && exists {
```
gonzolino: The error check should come before the exists check — standard Go error-first pattern. Currently `err == nil && exists` is checked first, then `err != nil` in a separate block. Inverting these would be more idiomatic and easier to read, and we could drop the `err == nil` and only keep `exists`.

internal/controllers/snapshot_controller.go
gonzolino: `openIroncoreImageSource` resolves the full OCI image source (network I/O, content stream initialization) just to extract `snapshotSize` and `digest`, then the `ReadCloser` is immediately discarded. The long-ops controller re-resolves the same source when it actually populates. Consider deferring size/digest extraction to the long-ops controller to avoid the redundant resolution.

internal/controllers/snapshot_longops_controller.go
```diff
@@ -0,0 +1,432 @@
...
+			defer wg.Done()
+			r.processFlattenQueue(ctx, log)
+		}()
+	}
+
+	wg.Wait()
+	return nil
+}
+
+func (r *SnapshotLongOpsReconciler) processPopulateQueue(ctx context.Context, log logr.Logger) {
```
gonzolino: Can we rename this function to `processNextPopulateWorkItem` to keep the naming consistent with existing controllers? Same for the Flatten function.

internal/controllers/snapshot_longops_controller.go
```diff
@@ -0,0 +1,432 @@
...
+
+		err := r.processPopulateOperation(ctx, snapshotID)
+		if err != nil {
+			log.Error(err, "populate operation failed", "snapshotId", snapshotID)
+			r.populateQueue.AddRateLimited(snapshotID)
+		} else {
+			r.populateQueue.Forget(snapshotID)
+		}
+		r.populateQueue.Done(snapshotID)
+	}
```
gonzolino: Can we keep the code structure of this function the same as `processNextWorkItem` from the existing controllers? And also move the for loop out to the caller in `Start`.

gonzolino (changes requested):

cmd/volumeprovider/app/app.go
```diff
@@ -173,6 +190,21 @@ func Run(ctx context.Context, opts Options) error {
 		setupLog.Error(err, "Worker size validation failed")
 		return err
 	}
+	if opts.Ceph.SnapshotPopulateLongOpsWorkerSize <= 0 {
+		err := fmt.Errorf("invalid configuration: snapshot-populate-long-ops-worker-size must be greater than 0, but got %d", opts.Ceph.SnapshotPopulateLongOpsWorkerSize)
```
kasabe28, line 196: Can we assign default worker size here, instead of returning an error? Same for other 2 workers.

internal/controllers/snapshot_longops_controller.go
```diff
@@ -0,0 +1,432 @@
...
+		if err := flattenImage(log, r.conn, pools[0], childImgs[0]); err != nil {
+			return fmt.Errorf("failed to flatten child %s: %w", childImgs[0], err)
+		}
+		r.flattenQueue.Add(snapshotID)
+		return nil
+	}
+
+	// No children remaining: perform cleanup first, then remove the finalizer.
+	rbdSnapshot := img.GetSnapshot(snapName)
+	if err := removeSnapshot(rbdSnapshot); err != nil {
```
kasabe28: Flatten operation is doing cleanup also, which I think should be there in existing snapshot controller. This controller should purely do flatten and image population only like `image_longops_controller`. Maybe we can add one more state `Flattened` once all child images are flattened and have a check for it in existing snapshot controller.

internal/controllers/snapshot_controller.go
```diff
@@ -312,6 +246,16 @@ func (r *SnapshotReconciler) reconcileSnapshot(ctx context.Context, id string) e
 		return fmt.Errorf("failed to reconcile snapshot: %w", err)
 	}
 
+	// For ironcore-image-backed snapshots, population is asynchronous. The long-ops controller
```
kasabe28: We can remove this block by moving line 259 -> snapshot.Status.State = providerapi.SnapshotStateReady into 'reconcileVolumeImageSnapshot()`

internal/controllers/snapshot_longops_controller.go
```diff
@@ -0,0 +1,432 @@
...
+
+	snapshot.Status.State = providerapi.SnapshotStateReady
+	if _, err := r.store.Update(ctx, snapshot); err != nil {
+		return fmt.Errorf("failed to update snapshot: %w", err)
+	}
+
+	return nil
+}
+
+func (r *SnapshotLongOpsReconciler) processFlattenOperation(ctx context.Context, snapshotID string) error {
```
kasabe28: I think we can handle this flatten operation in `image_longops_controller` only, by checking snapshot update event in `Start()` like `image_controller`, check for `Flattening` state, then get source details and add parent image ID in flattenQueue.

kasabe28 (commented):
Optional: `longops` from controller name can be replaced by actual long operation name like `flattening` or `population`.

cmd/volumeprovider/app/app.go
sujeet01: Defaults are already set in `Options.Defaults()`. We still error on `0` in `Run()` to catch bad config.

api/snapshot.go
sujeet01: Addressed.

internal/controllers/snapshot_controller.go
sujeet01: Addressed.

internal/controllers/snapshot_longops_controller.go
sujeet01: Done.

internal/controllers/snapshot_controller.go
sujeet01: Done.  `SnapshotStateReady` is now set and persisted in `reconcileVolumeImageSnapshot` only after a successful `createSnapshot`.  

Also fixed a prior behaviour where `Get` on a missing source image could result in the snapshot being marked `Ready`. A missing source now sets the snapshot to `Failed` state.

internal/controllers/snapshot_longops_controller.go
sujeet01: In `image_controller`, snapshot events are only used as a signal to requeue `images`. 
But our snapshot flattening needs snapshot context `snapshotID` + (`parentRBD`, `snapName`) and moving it would force `ImageLongOpsReconciler` to also own snapshot store/state, increasing coupling without reducing overall complexity.

internal/controllers/snapshot_longops_controller.go
sujeet01: Addressed.

sujeet01:
> I am wondering if it wouldn't be better to have proper controllers to do asynchronous operations instead of worker pools.
> 
> This way we would have a standard reconciler with all the stuff controller-runtime offers. This controller would watch Snapshot and filter for states "Populating" and "Flattening", handle the long-running operations and bring the state back to "Ready" once finished.
> 
> What do you think?

Current PR branch now includes the dedicated controller-based approach for the async operations.
I’ve also pushed my initial in-controller worker-pool approach to a separate branch [here](https://github.com/ironcore-dev/ceph-provider/compare/main...opensovereigncloud:ceph-provider:osc/enh/in-controller-async) for side-by-side comparison.

internal/controllers/snapshot_controller.go
```diff
@@ -291,92 +366,56 @@ func (r *SnapshotReconciler) reconcileSnapshot(ctx context.Context, id string) e
...
-	if snapshot.Labels != nil {
-		if arch, found := snapshot.Labels[providerapi.MachineArchitectureLabel]; found {
-			log.V(2).Info("Snapshot architecture", "architecture", arch)
-			platform = toPlatform(&arch)
+	if !slices.Contains(snapshot.Finalizers, SnapshotFinalizer) {
+		snapshot.Finalizers = append(snapshot.Finalizers, SnapshotFinalizer)
+		if _, err := r.store.Update(ctx, snapshot); err != nil {
+			return fmt.Errorf("failed to set finalizers: %w", err)
 		}
 	}
```
gonzolino: This finalizer check (lines 382-387) is unreachable dead code. The finalizer was already unconditionally set at lines 327-332 above, and by this point the code only arrives here for snapshots whose state is `Pending` (all other states return early). Every snapshot that reaches line 382 already has the finalizer from line 328.

internal/controllers/snapshot_controller.go
```diff
@@ -168,58 +212,84 @@ func (r *SnapshotReconciler) deleteSnapshot(ctx context.Context, log logr.Logger
...
+			return fmt.Errorf("failed to update snapshot: %w", err)
 		}
-	}()
+	}
 
-	if err := flattenChildImages(log, r.conn, img); err != nil {
-		return fmt.Errorf("failed to flatten snapshot child images: %w", err)
+	// Delegate flattening to the async flatten queue. The worker will set state to Flattened
+	// when flattening completes (or no children remain)
+	if len(childImgs) == 0 {
```
gonzolino: The code after this if clause does exactly the same as the code inside the if clause (`r.flattenQueue.Add(snapshot.ID); return nil`).

internal/controllers/snapshot_controller.go
sujeet01: Removed the duplicate check.

internal/controllers/snapshot_controller.go
sujeet01: Removed the redundant code. Sync delete now uses a single enqueue path and the async flatten worker handles the rest.

cmd/volumeprovider/app/app.go
gonzolino, line 1: Instead of having fixed worker sizes, I think it would be better to take over the mechanism from the oci.image.LocalCache from [provider-utils](https://github.com/ironcore-dev/provider-utils/blob/main/ociutils/oci/image.go):

The cache is [started in a goroutine](https://github.com/ironcore-dev/libvirt-provider/blob/main/cmd/libvirt-provider/app/app.go#L425). This goroutine then [listens on a channel in a loop](https://github.com/ironcore-dev/provider-utils/blob/main/ociutils/oci/image.go#L174) and reacts to [new work packages](https://github.com/ironcore-dev/provider-utils/blob/main/ociutils/oci/image.go#L182) (image pull requests in this case) or [notifies if work is finished](https://github.com/ironcore-dev/provider-utils/blob/main/ociutils/oci/image.go#L177) (image has been pulled). For each work package a [dedicated goroutine is spawned](https://github.com/ironcore-dev/provider-utils/blob/main/ociutils/oci/image.go#L196) that [reports back once the work is done](https://github.com/ironcore-dev/provider-utils/blob/main/ociutils/oci/image.go#L200). A [listener in the controller](https://github.com/ironcore-dev/libvirt-provider/blob/main/internal/controllers/machine_controller.go#L158) waits for the notifications and requeues the affected object so reconciliation can continue.

The benefit of this approach is that we only have a single idling goroutine when there is no work to do. When async work comes in, goroutines are dynamically scaled up as needed. Since everything is based on goroutines and channels for communication, the solution should also be lightweight. To make sure the number of parallel workers does not explode, we should add a config value for max parallel workers.

sujeet01: Implemented the suggested pattern in a separate PR https://github.com/ironcore-dev/ceph-provider/pull/876 for comparison and review.

PR #857, opened 2026-04-14
Enhance volume list label filter operation

balpert89:
# Proposed Changes

- Add `IRIVolume` labels to `Image` during volume creation
- Add `ListByLabels` function to list `omap` store images using label selector
- Remove `filterVolumes` function to enhance volume list processing

Fixes #851 

## Summary by CodeRabbit

* **New Features**
  * Faster label-based volume queries via an in-memory label index; selector-based listings are accelerated.
  * Image creation now preserves source volume labels on new images.

* **Bug Fixes**
  * More robust handling of missing storage entries to yield consistent “not found” responses.
  * More reliable handling of deletions when finalizers/tombstones are involved, improving event consistency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
-CONTROLLER_TOOLS_VERSION ?= v0.19.0
+CONTROLLER_TOOLS_VERSION ?= v0.17.3
 ADDLICENSE_VERSION ?= v1.2.0
 GOLANGCI_LINT_VERSION ?= v2.5
 #ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
-CONTROLLER_TOOLS_VERSION ?= v0.17.3
+CONTROLLER_TOOLS_VERSION ?= v0.19.0
 ADDLICENSE_VERSION ?= v1.2.0
 GOLANGCI_LINT_VERSION ?= v2.5
 #ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
-CONTROLLER_TOOLS_VERSION ?= v0.17.3
+CONTROLLER_TOOLS_VERSION ?= v0.19.0
 ADDLICENSE_VERSION ?= v1.2.0
 GOLANGCI_LINT_VERSION ?= v2.5
 #ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
-CONTROLLER_TOOLS_VERSION ?= v0.19.0
+CONTROLLER_TOOLS_VERSION ?= v0.17.3
 ADDLICENSE_VERSION ?= v1.2.0
 GOLANGCI_LINT_VERSION ?= v2.5
 #ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
--- a/internal/omap/omap.go
+++ b/internal/omap/omap.go
@@ -238,7 +238,11 @@ func (s *Store[E]) Delete(ctx context.Context, id string) error {
 	}
 
 	if len(obj.GetFinalizers()) == 0 {
-		return s.delete(ioCtx, id)
+		if err := s.delete(ioCtx, id); err != nil {
+			return err
+		}
+		s.removeFromLabelIndex(id, obj.GetLabels())
+		return nil
 	}
 
 	if obj.GetDeletedAt() != nil {
@@ -457,7 +461,7 @@ func (s *Store[E]) ListByLabels(ctx context.Context, labelSelector map[string]st
 		}
 
 		labelSelect = append(labelSelect, sizedLabel{
-			ids:  ids,
+			ids:  ids.Clone(),
 			size: ids.Len(),
 		})
 	}
@@ -476,7 +480,7 @@ func (s *Store[E]) ListByLabels(ctx context.Context, labelSelector map[string]st
 		ids := info.ids
 		if isFirstLabel {
 			// Use the smallest set to initialize the intersection (copy to avoid modifying the index set).
-			intersection = ids.Clone()
+			intersection = ids
 			isFirstLabel = false
 		} else {
 			// Intersect the current result with the next smallest set.
--- a/internal/omap/omap.go
+++ b/internal/omap/omap.go
@@ -8,10 +8,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
 	"github.com/ceph/go-ceph/rados"
+	"github.com/go-logr/logr"
 	utilssync "github.com/ironcore-dev/ceph-provider/internal/sync"
 	"github.com/ironcore-dev/ceph-provider/internal/utils"
 	apiutils "github.com/ironcore-dev/provider-utils/apiutils/api"
@@ -31,7 +33,7 @@ type Options[E apiutils.Object] struct {
 	CreateStrategy CreateStrategy[E]
 }
 
-func New[E apiutils.Object](conn *rados.Conn, pool string, opts Options[E]) (*Store[E], error) {
+func New[E apiutils.Object](conn *rados.Conn, pool string, log logr.Logger, opts Options[E]) (*Store[E], error) {
 	if conn == nil {
 		return nil, fmt.Errorf("must specify conn")
 	}
@@ -55,10 +57,13 @@ func New[E apiutils.Object](conn *rados.Conn, pool string, opts Options[E]) (*St
 		pool:     pool,
 		omapName: opts.OmapName,
 
-		watches: sets.New[*watch[E]](),
+		// Initialize the label index
+		labelIndex: make(map[string]sets.Set[string]),
+		watches:    sets.New[*watch[E]](),
 
 		newFunc:        opts.NewFunc,
 		createStrategy: opts.CreateStrategy,
+		log:            log,
 	}, nil
 }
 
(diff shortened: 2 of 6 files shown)


kasabe28:
@coderabbitai review

internal/omap/omap.go
```diff
@@ -68,14 +73,74 @@ type Store[E apiutils.Object] struct {
 	conn     *rados.Conn
 	pool     string
 	omapName string
+	log      logr.Logger
```
gonzolino: The logger addition to the store is partly needed for the label index helpers, but a lot of the log statements added throughout this PR (in `Delete`, `Update`, `delete`, `deleteOmapValue`, etc.) are general observability improvements unrelated to label indexing.

Instead of adding them in this PR, can we create a dedicated PR just for the logging enhancements?

internal/volumeserver/volume_list.go
```diff
@@ -8,98 +8,117 @@ import (
...
+	store.Store[E]
+	ListByLabels(ctx context.Context, labelSelector map[string]string) ([]E, error)
+}
+
+func (s *Server) getIriVolume(ctx context.Context, log logr.Logger, imageId string) (*iri.Volume, error) {
 	cephImage, err := s.imageStore.Get(ctx, imageId)
 	if err != nil {
-		if errors.Is(err, utils.ErrVolumeNotFound) {
-			return nil, fmt.Errorf("failed to get image %s: %w", imageId, utils.ErrVolumeNotFound)
+		if errors.Is(err, store.ErrNotFound) {
```
gonzolino, line 21: Switching the expected error types (`utils.ErrVolumeNotFound` to `store.ErrNotFound`) changes the error contract of this method.
I can't find the old error type (`utils.ErrVolumeNotFound`) in the map store, so I am wondering: Was this a bug before and this change is a fix unrelated to the label filtering?

If so, I would prefer opening a separate PR for it.

internal/volumeserver/volume_list.go
```diff
@@ -8,98 +8,117 @@ import (
...
 		}
-
 		iriVolume, err := s.convertImageToIriVolume(cephImage)
 		if err != nil {
-			return nil, err
+			log.Error(err, "Failed to convert ceph image to ori volume", "imageID", cephImage.GetID())
+			continue
 		}
-
 		res = append(res, iriVolume)
```
gonzolino, line 57: The old `listVolumes` returned an error if any single image failed `convertImageToIriVolume`. The new `listAndConvert` logs and continues. This is a change in behavior, that I think is unrelated to label index filtering.

I think I would prefer not bundling this change in this PR. If the old behavior is unexpected we should open a bug for it and fix it in a separate PR.

Makefile
```diff
@@ -156,7 +156,7 @@ GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
-CONTROLLER_TOOLS_VERSION ?= v0.17.3
+CONTROLLER_TOOLS_VERSION ?= v0.19.0
```
gonzolino: The controller-tools and ironcore dependency bumps are unrelated to the label index feature.
Lets splitt them out as their own PR.

internal/omap/omap.go
```diff
@@ -108,9 +173,11 @@ func (s *Store[E]) getSingleOmapValue(ioCtx *rados.IOContext, omapName, key stri
 
 func (s *Store[E]) deleteOmapValue(ioCtx *rados.IOContext, omapName, key string) error {
 	if err := ioCtx.RmOmapKeys(omapName, []string{key}); err != nil {
+		if errors.Is(err, rados.ErrNotFound) {
```
gonzolino: I don't think this check is needed. A caller using `errors.Is` to check for `rados.ErrNotFound` would still be able to detect it if the error is wrapped.

internal/omap/omap.go
```diff
@@ -234,25 +308,53 @@ func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
 		return utils.Zero[E](), err
 	}
 
+	// Begin OMAP Update Logic
+	deleted := false
+	oldLabels := oldObj.GetLabels()
+	newLabels := obj.GetLabels()
+
 	if obj.GetDeletedAt() != nil && len(obj.GetFinalizers()) == 0 {
+		s.log.V(1).Info("Update triggers physical deletion", "id", obj.GetID())
```
gonzolino: I am not sure I understand what these new changes are doing, and how they relate to the label filtering.

Can you elaborate a bit more?

gonzolino (commented):
There are some changes in this PR which I believe are not directly related to the label filtering.

Let's discuss if we can separate those changes from this PR.

internal/omap/omap.go
kasabe28: removed logging enhancements

internal/volumeserver/volume_list.go
kasabe28: I think it seems a bug. But as per comment I am removing this change from current PR.

internal/volumeserver/volume_list.go
kasabe28: Reverted.

internal/omap/omap.go
kasabe28: We are incrementing `resourceVersion` only when object is not marked for deletion. Also introduced new boolean `deleted` to add correct `eventType` and remove/update `labelIndex` accordingly.

internal/volumeserver/volume_list.go
```diff
@@ -8,98 +8,113 @@ import (
...
+				return &iri.ListVolumesResponse{Volumes: []*iri.Volume{}}, nil
 			}
-			return &iri.ListVolumesResponse{
-				Volumes: []*iri.Volume{},
-			}, nil
+			log.Error(err, "Error getting volume by ID", "VolumeID", filter.Id)
+			return nil, err
 		}
+		// Found by ID
+		return &iri.ListVolumesResponse{Volumes: []*iri.Volume{volume}}, nil
```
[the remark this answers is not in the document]
kasabe28: We don't have any such use case now

internal/volumeserver/volume_list.go
```diff
@@ -8,13 +8,20 @@ import (
...
 	"github.com/ironcore-dev/ceph-provider/api"
 	"github.com/ironcore-dev/ceph-provider/internal/utils"
 	iri "github.com/ironcore-dev/ironcore/iri/apis/volume/v1alpha1"
+	providerapi "github.com/ironcore-dev/provider-utils/apiutils/api"
 	"github.com/ironcore-dev/provider-utils/storeutils/store"
-	"k8s.io/apimachinery/pkg/labels"
 )
 
+// imageListerWithLabels type parameter E is constrained to be a providerapi.Object.
+type imageListerWithLabels[E providerapi.Object] interface {
```
gonzolino: I wonder if it would be better to extend the `List(ctx)` function in the `store.Store` interface at https://github.com/ironcore-dev/provider-utils/blob/main/storeutils/store/store.go#L50 to support some kind of filter parameter.

I think this would significantly reduce the amount of code needed for this change, and it might be useful to other store implementations / use-cases.

Let's discuss this on the next Community Call.

internal/omap/omap.go
gonzolino: This comment is not directly related to the label filtering, but it might address the motivation in the original issue #851

> "The number of volumes can have detrimental effects on the performance of the ceph-peovider to process volumes"

I think the main reason for the drop in performance on large numbers of volumes is the `iteratorSize` value of 10 here in the `ioCtx.GetAllOmapValues` call.

The `iteratorSize` defines how many objects are fetched from the Omap store in a single request. The client will do requests with this size until it has fetched all objects. This means if there are 10 objects in the Omap store a single request will fetch everything. If there are 100 objects it requires 10 requests, 1000 object require 100 requests, etc.

Therefore, in order to fix the performance issues mentioned in the issue, I would advice to make this value configurable and try to use larger values than 10 (e.g. 100). While I don;t have hard numbers I would expect this to significantly boost performance without adding caching complexities to the Omap store implementation.

internal/volumeserver/volume_list.go
balpert89: I agree, it makes sense to extend the `List` functionality in the provider-utils package.

gonzolino: I have a PR ready to extend `List` in provider-utils: https://github.com/ironcore-dev/provider-utils/pull/56

internal/omap/omap.go
kasabe28: I agree. So, we should decide `defaultIteratorSize` here.

gonzolino: After talking to some Ceph experts I got the guidance that 1000 should be a good default value for the omap iterator size.

I prepared a PR to set this and also provide a way to configure a different number: https://github.com/ironcore-dev/ceph-provider/pull/889

internal/volumeserver/volume_list.go
kasabe28: Updated PR to support List with labels and field indexes using modified provider-utils List()

internal/omap/omap.go
```diff
@@ -99,6 +108,183 @@ type Store[E apiutils.Object] struct {
...
+
+func formatLabel(key, value string) string {
+	return formatIndexKey(key, value)
+}
+
+func formatField(key, value string) string {
+	return formatIndexKey(key, value)
+}
+
+// updateLabelIndex updates the index for a single object based on its labels.
```
gonzolino: Comment does not match the function name

internal/omap/omap.go
```diff
@@ -99,6 +108,183 @@ type Store[E apiutils.Object] struct {
...
+	if err != nil && !errors.Is(err, rados.ErrNotFound) {
+		if errors.Is(err, rados.ErrNotFound) {
+			s.log.V(1).Info("OMAP not found during initial index initialization", "index", indexName)
+			return nil
+		}
+		s.log.Error(err, "Index initialization failed", "index", indexName)
+		return fmt.Errorf("failed to get all omap values for %s index initialization: %w", indexName, err)
+	}
+
+	for key := range index {
```
gonzolino: During initialization, shouldn't the index always be empty? I don't think we need to "clean it up" here.

internal/omap/omap.go
```diff
@@ -99,6 +108,183 @@ type Store[E apiutils.Object] struct {
...
+
+	labelIndexMu sync.RWMutex
+	labelIndex   map[string]sets.Set[string] // Add label index field (labelKey=labelValue -> Set[objectID])
+	fieldIndexMu sync.RWMutex
+	fieldIndex   map[string]sets.Set[string] // Add field index field (fieldName=fieldValue -> Set[objectID])
+
+}
+
+// --- Internal Index Helpers ---
+func formatIndexKey(key, value string) string {
```
gonzolino: How will index keys in this format work with [`HasLabel`](https://github.com/ironcore-dev/provider-utils/blob/main/storeutils/store/listoptions.go#L30) queries? Such queries use an [`Exists` operator](https://github.com/ironcore-dev/provider-utils/blob/main/storeutils/store/listoptions.go#L33) on the key, without caring for the value. This index key format expects all index keys to consist of a key + value combination.

internal/omap/omap.go
```diff
@@ -99,6 +108,183 @@ type Store[E apiutils.Object] struct {
...
+
+func (s *Store[E]) initializeIndex(indexName string, index map[string]sets.Set[string], indexBuilder func(E) map[string]string) error {
+	ioCtx, err := s.conn.OpenIOContext(s.pool)
+	if err != nil {
+		return fmt.Errorf("failed to open IO context for %s index initialization: %w", indexName, err)
+	}
+	defer ioCtx.Destroy()
+
+	omapValues, err := ioCtx.GetAllOmapValues(s.omapName, "", "", s.iteratorSize)
+	if err != nil && !errors.Is(err, rados.ErrNotFound) {
```
gonzolino:
```suggestion
	if err != nil {
```

gonzolino (commented):
Still reviewing, but here are some comments already.

PR #821, opened 2026-01-14
Separate API into `Volume`, `Image` and `Snapshot`

lukasfrank:
Scaffold to illustrate a simpler and maintainable flow on how to realize snapshots. The orchestration is done in  a `Volume` resource to have a 1:1 mapping with `Image` and `Snapshot` to the ceph core resources.

PR #876, opened 2026-06-22
Add bounded async runner for long running ops

### LINKED ISSUE #822: Move long-running ceph operations out of controllers

sujeet01:
# Proposed Changes

This PR moves long-running snapshot and image work (IronCore populate, flatten-on-delete, child-flatten) to a bounded async runner so reconcile workers are not blocked. A shared runner runs these long operations with capped concurrency (--async-max-workers, default 10), deduplicates by key, and requeues reconcilers when async work finishes. 

Fixes #822 
Ref https://github.com/ironcore-dev/ceph-provider/pull/837#discussion_r3387738530

## Release Notes

* **New Features**
  * Added asynchronous deletion processing for images by deferring child flattening, with a new “FlatteningChildren” state (shown as pending where applicable).
  * Added asynchronous snapshot lifecycle support, including “Populating” and “Flattening/Flattened” states.
  * Introduced `--async-max-workers` (default: 10) to control async concurrency.

* **Chores**
  * Updated Kubernetes/controller-runtime, OpenTelemetry, Prometheus, and other dependencies.

* **Tests**
  * Adjusted integration test suite async worker configuration to align with the new default.


gonzolino (commented):
(sorry, clicked on the comment button by mistake 🙈)

internal/async/runner.go
```diff
@@ -0,0 +1,213 @@
...
+func New(log logr.Logger, name string, maxWorkers int) *Runner {
+	if maxWorkers <= 0 {
+		maxWorkers = 5
+	}
+	return &Runner{
+		name:       name,
+		log:        log.WithName("async-runner").WithValues("runner", name),
+		maxWorkers: maxWorkers,
+		submitCh:   make(chan submitRequest),
+		doneCh:     make(chan DoneEvent),
```
gonzolino, line 87: I think `doneCh` needs to be buffered to give all workers a chance to finish their current work in `run(ctx context.Context, key string, op Operation)`.
```suggestion
		doneCh:     make(chan DoneEvent, maxWorkers),
```

internal/async/runner.go
```diff
@@ -0,0 +1,213 @@
...
+		select {
+		case <-ctx.Done():
+			return
+
+		case evt := <-r.doneCh:
+			delete(inFlight, evt.Key)
+			if active > 0 {
+				active--
+			} else {
+				r.log.Info("unexpected done event with active counter already zero",
```
gonzolino, line 136: Under normal operation this `else` block should never execute. I think if it happens, it should throw a proper error and not write an info-level log.

internal/async/runner.go
```diff
@@ -0,0 +1,213 @@
...
+					"key", evt.Key, "err", evt.Err)
+			}
+			r.notify(evt)
+
+		case req := <-r.submitCh:
+			switch {
+			case func() bool { _, ok := inFlight[req.key]; return ok }():
```
gonzolino, line 143:
```suggestion
			_, alreadyRunning := inFlight[req.key]
			switch {
			case alreadyRunning:
```

internal/async/runner.go
gonzolino, line 1: I think we should add some tests to verify the behavior of the runner.

internal/async/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	doneCh   chan DoneEvent
+}
+
+func New(log logr.Logger, name string, maxWorkers int) *Runner {
+	if maxWorkers <= 0 {
+		maxWorkers = 5
+	}
+	return &Runner{
+		name:       name,
+		log:        log.WithName("async-runner").WithValues("runner", name),
```
gonzolino, line 84: Personally, I would not add config to the `log` here, as this can be set by the caller.


```suggestion
		log:        log,
```

internal/async/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	running   bool
+	listeners []Listener
+
+	submitCh chan submitRequest
+	doneCh   chan DoneEvent
+}
+
+func New(log logr.Logger, name string, maxWorkers int) *Runner {
+	if maxWorkers <= 0 {
+		maxWorkers = 5
```
gonzolino, line 80: I would prefer an error for an invalid `maxWorkers` value. Just silently setting it to 5 might be confusing to users if there is a misconfiguration.

internal/controllers/image_async.go
```diff
@@ -0,0 +1,66 @@
...
+		return fmt.Errorf("failed to open image: %w", err)
+	}
+	defer closeImage(log, img)
+
+	pools, childImgs, err := img.ListChildren()
+	if err != nil {
+		return fmt.Errorf("failed to list children: %w", err)
+	}
+
+	for i, child := range childImgs {
```
gonzolino, line 58: `processImageFlattenOperation` is run inside an `async.Runner` and can therefore run asynchronously, with multiple operations running in parallel.

But then, it will flatten all child images inside a for loop, one after the other. I think it would be more efficient to asynchronously flatten the child images, thereby having more parallelization in the flatten process.

internal/controllers/image_async.go
```diff
@@ -0,0 +1,66 @@
...
+		log.V(2).Info("Flattening child image", "child", child, "remaining", len(childImgs)-i)
+		if err := flattenImage(log, r.conn, pools[i], child); err != nil {
```
gonzolino, line 60: `flattenImage` uses `r.conn` to create its own IOContext. We could instead reuse the IOContext created above on line 36.

internal/controllers/snapshot_async.go
```diff
@@ -0,0 +1,369 @@
...
+	if s.DeletedAt != nil || s.Status.State != providerapi.SnapshotStatePopulating || s.Source.IronCoreImage == "" {
+		return nil, false, nil
+	}
+	return s, true, nil
+}
+
+// processFlattenOperation is the async operation submitted to the runner during snapshot
+// deletion. It flattens all child images of the snapshot and then moves the snapshot to the
+// Flattened state so reconciliation can run the terminal cleanup.
+func (r *SnapshotReconciler) processFlattenOperation(ctx context.Context, snapshotID string) error {
```
gonzolino, line 268: Can we somehow share the flatten code for snapshots and images?

internal/controllers/snapshot_async.go
```diff
@@ -0,0 +1,369 @@
...
+	snapshot, stillPopulating, err := r.syncSnapshotFromStore(ctx, snapshot.ID)
+	if err != nil {
+		return err
+	}
+	if !stillPopulating {
+		return nil
+	}
+
+	// Create the backing RBD image if needed.
+	rbdImg, err := openImage(ioCtx, imageName)
```
gonzolino, line 174: `rbdImg` is not closed after usage/

gonzolino (changes requested):
I focused on changes in internal/async/runner.go and the two new async files in internal/controllers.

Also: I will try to update PR #821 this or next week to simplify our data model in ceph-provider, and thereby our controller logic as well. I believe this should also simplify the async code related to snapshots.

PR #881, opened 2026-07-01
Fix `multi-arch` image population

### LINKED ISSUE #880: Multi-arch image population is not deterministic if no architecture hint is given
**Describe the bug**
Multi-arch image population is not deterministic if no architecture  hint is given. `ironcore-image` resolves the first layer which is error prone.

lukasfrank:
# Proposed Changes

- Fix `multi-arch` image population by emitting an event and stop populating a random layer

Fixes #880 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image reference resolution by emitting a warning event when the requested platform has no matching platform.
  * Fixed image creation from volume image data sources to set `ImageArchitecture` directly from the provided architecture field.

* **Chores**
  * Updated Go module dependencies, including several `golang.org/x/*` libraries and related indirect packages, to newer versions.


internal/controllers/image_controller.go
```diff
@@ -440,6 +441,9 @@ func (r *ImageReconciler) reconcileSnapshot(ctx context.Context, log logr.Logger
 
 	resolvedImg, err := osImgSrc.Resolve(ctx, img.Spec.Image)
 	if err != nil {
+		if errors.Is(err, remote.ErrNoPlatformMatch) {
+			r.Eventf(img.Metadata, corev1.EventTypeWarning, "NoPlatformMatch", "No platform match: %v", err)
```
gonzolino: Maybe we change the description a bit. Currently `"NoPlatformMatch` and `"No platform match: %v"` are kind of redundant.
```suggestion
			r.Eventf(img.Metadata, corev1.EventTypeWarning, "NoPlatformMatch", "Image %s has no matching platform: %v", err)
```

internal/volumeserver/volume_create.go
```diff
@@ -109,6 +100,9 @@ func (s *Server) createImageFromVolume(ctx context.Context, log logr.Logger, vol
 			if imageSize == 0 {
 				return nil, fmt.Errorf("must specify size when creating volume from image data source")
 			}
+			if dataSource.ImageDataSource.Architecture != "" {
+				volArch = ptr.To(dataSource.ImageDataSource.Architecture)
```
gonzolino, line 104: The new code handles `volume.Spec.VolumeDataSource.ImageDataSource`, but ignores the old deprecated `volume.Spec.Image`.

I am not sure if we can fix this for the deprecated field, as there is currently no architecture field for it. But maybe we should emit a hint or something...

lukasfrank: You are right. I propose to tackle this issue next to clean this up https://github.com/ironcore-dev/ironcore/issues/1495

What about that?

gonzolino: Sounds good 👍

internal/controllers/image_controller.go
lukasfrank: fixed

gonzolino (approved):

PR #888, opened 2026-07-06
Fetch omap values by keys instead of fetching all values

### LINKED ISSUE #887: Optimize omap Store single-key lookups
# Summary

In the omap store, replace the usage of `GetAllOmapValues` ([docs](https://pkg.go.dev/github.com/ceph/go-ceph@v0.40.0/rados#IOContext.GetAllOmapValues)) with `GetOmapValuesByKeys` ([docs](https://pkg.go.dev/github.com/ceph/go-ceph@v0.40.0/rados#ReadOp.GetOmapValuesByKeys)) if only specific omap values are to be fetched.

## Motivation

The omap store currently uses `GetAllOmapValues(omapName, "", key, 10)` to fetch single values from the omap store. This is a workaround, where all omap values are scanned and the `key` is used as a prefix filter.

This implementation has issues
1. `key` could potentially be a prefix for a different omap value, leading to unwanted matches in the result
2. `GetAllOmapValues` iterates through all omap values with a batch size of 10, which is inefficient, especially with lots of elements in omap.
3. The solution only works for single keys. To fetch multiple keys it is required to call `GetAllOmapValues` once per key.

Switching to `GetOmapValuesByKeys(keys []string)` ([docs](https://pkg.go.dev/github.com/ceph/go-ceph@v0.40.0/rados#ReadOp.GetOmapValuesByKeys)) will improve performance for large omaps and will make it easier in the future to efficiently fetch multiple keys (e.g. to implement indexing in the omap store).

gonzolino:
# Proposed Changes

- Remove `getSingleOmapValue` method from omap store
- Add `getOmapValuesByKeys` method to omap store

`getSingleOmapValue` was using [`GetAllOmapValues`](https://pkg.go.dev/github.com/ceph/go-ceph@v0.40.0/rados#IOContext.GetAllOmapValues) under the hood, (see #887 for why this is bad). `getOmapValuesByKeys` uses [`GetOmapValuesByKeys`](https://pkg.go.dev/github.com/ceph/go-ceph@v0.40.0/rados#ReadOp.GetOmapValuesByKeys) instead, which should perform better and allow better for future development (e.g. indexing).

Fixes #887

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing records so “not found” outcomes are more consistent across lookup paths.
  * Reduced inconsistencies in how lookup errors are reported.
  * Updated record retrieval to use a more efficient bulk-read strategy for key lookups, enhancing performance and reliability.

### DIFF
--- a/internal/omap/omap.go
+++ b/internal/omap/omap.go
@@ -99,14 +99,14 @@ func getOmapValuesByKeys(ioCtx *rados.IOContext, omapName string, keys []string)
 	step := op.GetOmapValuesByKeys(keys)
 
 	if err := op.Operate(ioCtx, omapName, rados.OperationNoFlag); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("ceph read operation failed: %w", err)
 	}
 
 	result := make(map[string][]byte, len(keys))
 	for {
 		kv, err := step.Next()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to iterate over ceph read results: %w", err)
 		}
 		if kv == nil {
 			return result, nil
--- a/internal/omap/omap.go
+++ b/internal/omap/omap.go
@@ -92,18 +92,27 @@ func (s *Store[E]) watchHandlers() []*watch[E] {
 	return s.watches.UnsortedList()
 }
 
-func (s *Store[E]) getSingleOmapValue(ioCtx *rados.IOContext, omapName, key string) ([]byte, error) {
-	omap, err := ioCtx.GetAllOmapValues(omapName, "", key, 10)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/omap/omap.go
```diff
@@ -99,18 +92,27 @@ func (s *Store[E]) watchHandlers() []*watch[E] {
...
 	}
 
-	value, ok := omap[key]
-	if !ok {
-		return nil, rados.ErrNotFound
+	result := make(map[string][]byte, len(keys))
+	for {
+		kv, err := step.Next()
+		if err != nil {
+			return nil, err
```
lukasfrank: It might be useful to wrap this error for debugging purposes

lukasfrank (approved):

